PR #49, opened 2024-06-15
Issue/57007697 configuration variables

NuclearTea:
Telecommands for configuration variables

Work In-Progress


DeflateAwning:
Is it okay if we delete this branch? Do you have everything you need out of it?

PR #62, opened 2024-06-19
Update arg len, correct indexing

NuclearTea:


DeflateAwning:
Please resubmit with only your changes. If we want to run an autoformatter, this needs to be configured with a shared plugin, and GitHub Action verification; otherwise, the git history/blames get messed up, and the merging/rebasing becomes a chore.

Issue #61, opened 2024-06-19
BUG - args_str final char is space, should be null


NuclearTea:
Repro Steps:

1. Start up program with UART Terminal ready to go (in debug mode)
2. Find a telecommand takes a parameter and place a breakpoint in that functions first line
4. Execute telecommand with parameter
5. Upon hitting breakpoint, inspect value of args_str (ideally inspecting binary)

Expected Behaviour:
the string of args (args_str) should terminate with a null.

Ex: CTS1+some_func(Bob) -> args_str = "Bob/0"

PR #64, opened 2024-06-20
Update Null terminator position for `arg_str`

NuclearTea:
#61 

Ensured that function without params (hello world), function with integer params (echo back uint32 args) and function with string params work as intended

### DIFF
--- a/firmware/Core/Src/rtos_tasks/rtos_tasks.c
+++ b/firmware/Core/Src/rtos_tasks/rtos_tasks.c
@@ -125,7 +125,7 @@ void TASK_handle_uart_telecommands(void *argument) {
 
 			// TODO: maybe log/print args too
 			
-			const uint8_t arg_len = end_of_args_idx - start_of_args_idx - 1;
+			const uint16_t arg_len = end_of_args_idx - start_of_args_idx - 1;
 			uint8_t args_str_no_parens[arg_len + 1];
 			memcpy(args_str_no_parens, &latest_tcmd[start_of_args_idx + 1], arg_len);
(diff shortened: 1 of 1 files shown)


DeflateAwning:
Fixes #61

firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -121,9 +125,10 @@ void TASK_handle_uart_telecommands(void *argument) {
 
 			// TODO: maybe log/print args too
 			
-			uint8_t args_str_no_parens[end_of_args_idx - start_of_args_idx + 2];
-			memcpy(args_str_no_parens, &latest_tcmd[start_of_args_idx+1], end_of_args_idx - start_of_args_idx - 1);
-			args_str_no_parens[end_of_args_idx - 1] = '\0';
+			const uint8_t arg_len = end_of_args_idx - start_of_args_idx - 1;
```
DeflateAwning, line 128: `arg_len` may overflow. `uint8_t` is not long enough.

DeflateAwning (changes requested):
Perfect other than the int size issue

firmware/Core/Src/rtos_tasks/rtos_tasks.c
NuclearTea: Made uint16, max value is ~65000, I think we won't have an argument that long 😄

DeflateAwning (approved):
Looks good to me. Thanks for fixing that! Please squash and merge!

Issue #46, opened 2024-06-14
LFS: Review the TODOs/FIXMEs in the `flash_driver.c` implementation to avoid hanging forever if flash module is unavailable


DeflateAwning:
* In the `flash_driver.c` implementation, there are some `while` loops that hang forever if the module is not available, or if writing is failing.
* Need to use a timeout mechanism that allows them to only run for a maximum number of milliseconds.
* Need to determine what that maximum number of milliseconds is.

PR #65, opened 2024-06-21
Added loop timeouts and attended to other FIXME comments in `flash_driver.c` (Issue #46)

Saksham-P:
## Issue #46
`LOOP_TIMEOUT_MS` defines the number of milliseconds to wait before we consider it a failed attempt or timeout.
Added timeout on all `// TODO:` sections with while loop. Fixed the bug where chip select would deactivate instead of activate in `FLASH_write_disable()`. 

### Possible next issues to consider:
1. For organizational purposes, we can create enum for errors as many errors are quite common in `flash_driver.c`
2. Go through the functions and compare them with the commands within the datasheet to make sure all errors are being handled and edge cases are being covered.

### DIFF
--- a/firmware/Core/Src/littlefs/flash_driver.c
+++ b/firmware/Core/Src/littlefs/flash_driver.c
@@ -7,8 +7,24 @@
 /// Timeout duration for HAL_SPI_READ/WRITE operations.
 #define FLASH_HAL_TIMEOUT_MS 5
 
-/// Duration to wait for the status register to update before timing out.
-#define FLASH_LOOP_TIMEOUT_MS 10
+// The following timeout values are sourced from Section 11.3.1, Table 56: "CFI system interface string"
+// and are interpreted using: https://www.infineon.com/dgdl/Infineon-AN98488_Quick_Guide_to_Common_Flash_Interface-ApplicationNotes-v05_00-EN.pdf
+
+/// Duration to wait for the status register to show that Write Enable Latch changed status.
+// Assume this value is the same as the typical timeout for a register change.
+#define FLASH_LOOP_REGISTER_CHANGE_TIMEOUT_MS 10
+
+/// Duration to wait for the status register to show that Write In Progress bit changed status, for writes.
+// "Typical timeout for page program" = 2^(0x09) us = 0.5 ms
+// "Max timeout for page program" = 2^(0x02) * typical = 2 ms
+// With a safety factor of 5, we get 10 ms.
+#define FLASH_LOOP_WRITE_TIMEOUT_MS 10
+
+/// Duration to wait for the status register to show that Write In Progress bit changed status, for erases.
+// "Typical timeout for sector erase" = 2^(0x09) ms = 512 ms
+// "Max timeout for sector erase" = 2^(0x03) * typical = 4096 ms
+// 4 seconds is a very long time, so no safety factor is applied here (to avoid lockups).
+#define FLASH_LOOP_SECTOR_ERASE_TIMEOUT_MS 4096
 
 // -----------------------------FLASH DRIVER FUNCTIONS-----------------------------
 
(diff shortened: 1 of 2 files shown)


DeflateAwning:
Fantastic comments on the "next issues". Go ahead and open individual issues for those!

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -135,31 +140,37 @@ uint8_t FLASH_write_enable(SPI_HandleTypeDef *hspi, uint8_t chip_number)
...
+            return 2;
+        }
+
         const uint8_t read_status_result = FLASH_read_status_register(hspi, chip_number, status_reg_buffer);
         if (read_status_result != 0) {
             FLASH_deactivate_chip_select();
-            return 1;
+            return 3;
         }
         wel = status_reg_buffer[0] & 2;
```
DeflateAwning, line 169: 2 is a magic number here. Why is it 2? Please write as `0x02`, also.

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -201,11 +212,17 @@ uint8_t FLASH_erase(SPI_HandleTypeDef *hspi, uint8_t chip_number, lfs_block_t ad
     uint8_t status_reg_buffer[1] = {0};
 
     // Keep looping as long as device is busy
-    // FIXME: needs a timeout
+    uint32_t loop_limit = HAL_GetTick() + LOOP_TIMEOUT_MS;
     uint8_t wip = 1;
     uint8_t err = 0;
```
DeflateAwning, line 217: Move `err` declaration inside the loop.

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -271,11 +284,16 @@ uint8_t FLASH_write(SPI_HandleTypeDef *hspi, uint8_t chip_number, lfs_block_t ad
...
     while (wip)
     {
+        uint32_t curren_time = HAL_GetTick();
+        if (curren_time > loop_limit) {
+            DEBUG_uart_print_str("Timeout while setting Write Enable Latch to 0\n");
```
DeflateAwning, line 294: Change all these error message to be more like: `"Flash write timeout\n"` (that is, explicitly say that a timeout happened, while doing a flash action, and that the flash action attempted was writing). If there's only one place it can timeout while doing a flash write, the detail about it being related to the Write Enable Latch isn't overly helpful for the system as-a-whole.

DeflateAwning (changes requested):
Very awesome overall! Glad to have these in. Some feedback on variable scope/types, comments, and errror strings. 

Have you tested each of these actions by removing the CS line on your STM32 board? If not, could you please do that and indicate that they all work as expected? I won't have time to check them for a while.

Saksham-P:
> Have you tested each of these actions by removing the CS line on your STM32 board? If not, could you please do that and indicate that they all work as expected?

Not sure what you mean by removing the CS line. But I did test the `FLASH_write_enable()`, `FLASH_write_disable()` with the `flash_write_hex()` telecommand with no errors, and after`FLASH_write_enable()` the Status Register value was `02`, after `FLASH_write_disable()` it was `00`, so enable and disable for sure work. 

Not entirely sure about write and read haven't fully tested those yet.

DeflateAwning:
Good stuf overall. Just occured to me that these are the textbook use case of a do-while loop (instead of just a while loop).

We want to make sure that the timeout mechanism indeed works if the flash module becomes defective, or if it becomes disconnected. The CS (chip select) wire is one of the wires between the flash module and the OBC/dev kit; it's what enables the flash module so it knows that it's the active device on the SPI bus. 

Recall that SPI buses can have multiple subnode devices (e.g., memory modules) connected to a single bus, so the CS line/wire is what specifies which device is being communicated with at a given time.

DeflateAwning:
In my testing, I made a couple changes as I noticed them. Figured it'd be easier to just commit than to explain; hopefully that's okay.

My big changes were:
1. Restructure the loops so that the timeout check isn't performed until the end of the loop. This avoids timing out if the loop iteration was successful right at the time limit. It also allows the loop to return faster, especially if the success is on the 0th iteration.
2. Refactored the SR1 status register byte masks to constants.
3. Fixed the `FLASH_erase()` function: the bit mask was wrong, and it was checking the "program error" byte of the SR1 (status register 1), instead of the "erase error" byte. 

I'd appreciate it if you could do a quick review on these changes and make sure you're happy with them. Then, feel free to merge!

DeflateAwning (approved):

Saksham-P:
The changes are amazing! I am wondering though, should we consider do while loops since the loop is being iterated at least once each time? Regardless it won't make a huge difference.

I will merge this pull request.

PR #71, opened 2024-06-24
Increase FLASH_HAL_TIMEOUT_MS to 10 ms

### LINKED ISSUE #70: LFS_ASSERT failure
LFS_mount() fails at

```
Line 119        LFS_ASSERT(err <= 0);
```

in `littlefs/lfs.c`. This appears to arise from the FLASH_... read function returning error values greater than zero. The above assertion failure was in response to this error in `flash_driver.c`:

```
Line 381    const HAL_StatusTypeDef rx_result_1 = HAL_SPI_Receive(hspi, (uint8_t *)rx_buffer, rx_buffer_len, FLASH_HAL_TIMEOUT_MS);
    if (rx_result_1 != HAL_OK) {
        return 3;
    }
```

This is from the main branch as updated yesterday with commit `407a78fab48120e8c2c39`.

JohnathanBurchill:
Fixes #70 failed read from flash chip.

Does not fix ASSERT issue where FLASH-related function return values are positive, but ifs-related functions assert errors to be less than 0. Suggest to open a new issue to address this specifically.

### DIFF
--- a/firmware/Core/Src/littlefs/flash_driver.c
+++ b/firmware/Core/Src/littlefs/flash_driver.c
@@ -5,7 +5,7 @@
 #include "debug_tools/debug_uart.h"
 
 /// Timeout duration for HAL_SPI_READ/WRITE operations.
-#define FLASH_HAL_TIMEOUT_MS 5
+#define FLASH_HAL_TIMEOUT_MS 10 
 
 // The following timeout values are sourced from Section 11.3.1, Table 56: "CFI system interface string"
 // and are interpreted using: https://www.infineon.com/dgdl/Infineon-AN98488_Quick_Guide_to_Common_Flash_Interface-ApplicationNotes-v05_00-EN.pdf


DeflateAwning:
Ooo that was the issue you were suggesting (positive vs. negative errors); indeed would be best to return negative error codes.

DeflateAwning (approved):
Thanks for identifying this issue and fixing it! 

I do wonder what's causing that overhead; shouldn't be an issue for now (I don't think), although may be an easy starting point to explore in the future if we need more performance out of the raw flash operations.

DeflateAwning:
No change required (I can add it later), but just as a community-made software project, the following comment can go a long way in documenting the "lesson" learned here, so that no one goes and accidentally decreases that seemingly-arbitrary number.

```
// Note: FLASH_read_data has sporadic timeouts at 5ms; 10ms is a safe bet.
// 512 bytes should take 2ms at 2Mbps.
// TODO: ^ investigate the HAL_SPI_Receive overhead (2ms expected, >5ms observed)
```

PR #76, opened 2024-06-26
Change `args_str` type from `uint8_t*` to `char*`

parker-research:
In order to follow the accepted convention (`uint8_t *` denotes a byte array with a known length, while `char *` represents a null-terminated c-string), the telecommand `args_str` argument type is changed from `const uint8_t *` to `const char *`.

This is a change which affects lots of lines of code, but is ultimately a tiny semantic change.

### DIFF
--- a/firmware/Core/Src/rtos_tasks/rtos_tasks.c
+++ b/firmware/Core/Src/rtos_tasks/rtos_tasks.c
@@ -72,6 +72,7 @@ void TASK_handle_uart_telecommands(void *argument) {
 			
 			// copy the buffer to the latest_tcmd buffer
 			latest_tcmd_len = UART_telecommand_buffer_write_idx;
+			// FIXME: rewrite the following memcpy/memset without the casts (which disregards the volatile qualifier)
 			memcpy(latest_tcmd, (uint8_t*) UART_telecommand_buffer, UART_telecommand_buffer_len); // copy the whole buffer to ensure nulls get copied too
 
 			// clear the buffer and reset the write pointer
@@ -141,7 +142,7 @@ void TASK_handle_uart_telecommands(void *argument) {
 			// TODO: maybe log/print args too
 			
 			const uint16_t arg_len = end_of_args_idx - start_of_args_idx - 1;
-			uint8_t args_str_no_parens[arg_len + 1];
+			char args_str_no_parens[arg_len + 1];
 			memcpy(args_str_no_parens, &latest_tcmd[start_of_args_idx + 1], arg_len);
 			args_str_no_parens[arg_len] = '\0';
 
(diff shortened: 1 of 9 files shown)


firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -72,6 +72,7 @@ void TASK_handle_uart_telecommands(void *argument) {
 			
 			// copy the buffer to the latest_tcmd buffer
 			latest_tcmd_len = UART_telecommand_buffer_write_idx;
+			// FIXME: rewrite the following memcpy/memset without the casts (which disregards the volatile qualifier)
```
JohnathanBurchill, line 75: Any idea of the significance of this FIXME, or how it might be done?

parker-research: You can't call memcpy nor memset with `volatile uint8_t*` (nor char) args because the implementation of those functions may use multi-byte operations under the hood. As a hack, I simply type-casted to make them non-volatile, but the correct approach is to use a for-loop.

The idea is that in the very-rare case that the ISR that writes to the `UART_...buffer` variable is called during this clearing operation, doing a mem-set is a for-loop is at least defined behaviour. 

Hardly matters since no UART telecommands will be issued in space, but type casts in C should be used sparingly. In case anyone uses with as an example, it'd be ideal to not cast

DeflateAwning (dismissed):

firmware/Core/Src/rtos_tasks/rtos_tasks.c
JohnathanBurchill: Thanks!

JohnathanBurchill:
There are now redundant casts in several telecommand/* .c files in the form `(char*)args_str`. The casts can be removed. :)

JohnathanBurchill:
Happy to see this change.

parker-research:
Ah perfect, thanks! Totally forgot that's why I did this in the first place. Will fix those up!

JohnathanBurchill (approved):
Looks great! I've compiled and flashed to the dev kit, flash_read_hex(0,0,100) works as expected.

I get one warning: 

```
Core/Src/telecommands/flash_telecommand_defs.c: In function 'TCMDEXEC_flash_write_hex':
Core/Src/telecommands/flash_telecommand_defs.c:165:10: warning: variable 'hex_str' set but not used [-Wunused-but-set-variable]
  165 |     char hex_str[max_num_bytes * 2 + 1];
      |          ^~~~~~~
```

Maybe that can be cleaned up in a separate PR, as it's been in `main` for some time.

PR #79, opened 2024-06-29
Scan i2c bus command

### LINKED ISSUE #2: Add I2C scan telecommands
Reference the I2C scanner example in the `debug_tools/debug_i2c.c` file.

* Argument 1: Which I2C port to scan, as an integer

parkerg-13:
Fixes #2

### DIFF
--- a/firmware/Core/Inc/rtos_tasks/rtos_tasks.h
+++ b/firmware/Core/Inc/rtos_tasks/rtos_tasks.h
@@ -5,6 +5,8 @@
 #include "main.h"
 #include "rtos_tasks/rtos_task_helpers.h"
 
+#define TCMD_MAX_RESPONSE_BUFFER_LENGTH 1024
+
 void TASK_DEBUG_print_heartbeat(void *argument);
 
 void TASK_handle_uart_telecommands(void *argument);
--- a/firmware/Core/Inc/telecommands/i2c_telecommand_defs.h
+++ b/firmware/Core/Inc/telecommands/i2c_telecommand_defs.h
@@ -1,0 +1,12 @@
+#ifndef __INCLUDE_GUARD__I2C_TELECOMMAND_DEFS_H__
+#define __INCLUDE_GUARD__I2C_TELECOMMAND_DEFS_H__
+
+#include <stdint.h>
+#include "telecommands/telecommand_definitions.h"
+
+
+
+uint8_t TCMDEXEC_scan_i2c_bus(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len);
+
+#endif /* __INCLUDE_GUARD__I2C_TELECOMMAND_DEFS_H__*/
--- a/firmware/Core/Src/telecommands/i2c_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/i2c_telecommand_defs.c
@@ -1,0 +1,72 @@
+#include "i2c_telecommand_defs.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "main.h"
+
+uint8_t TCMDEXEC_scan_i2c_bus(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+
+    I2C_HandleTypeDef* hi2c;
+    uint8_t bus_to_scan = atoi(args_str);
+
+    // Figure out which bus we want to scan
+    switch(bus_to_scan){
+        case 1:
+            hi2c = &hi2c1;
+            break;
+        case 2:
+            hi2c = &hi2c2;
+            break;
+        case 3: 
+            hi2c = &hi2c3;
+            break;
+        case 4:
+            hi2c = &hi2c4;
+            break;
+        default:
+            hi2c = NULL;
+            snprintf(response_output_buf, response_output_buf_len, "Enter a valid I2C bus!\n");
+            return 1;
+    }
+
+	char msg[16];
+    HAL_StatusTypeDef i2c_device_status;
+
+	// Go through all possible i2c addresses
+	for (uint8_t i = 0; i < 128; i++) {
+
+        i2c_device_status = HAL_I2C_IsDeviceReady(hi2c, (uint16_t)(i<<1), 3, 5);
+
+        // calculate remaining space in buffer. -1 for null terminator
+        size_t remaining_space = response_output_buf_len - strlen(response_output_buf) - 1;
+
+		if (i2c_device_status == HAL_OK) {
+			// O for OK
+            snprintf(msg, sizeof(msg), " 0x%02x ", i);
+		} 
+        else if (i2c_device_status == HAL_ERROR) {
+            // E for Error
+            snprintf(msg, sizeof(msg), "  _E_ ");
+        }
+        else if (i2c_device_status == HAL_TIMEOUT) {
+            // T for timeout
+            snprintf(msg, sizeof(msg), "  _T_ ");
+		}
+        else if (i2c_device_status == HAL_BUSY) {
+            // B for busy
+            snprintf(msg, sizeof(msg), "  _B_ ");
+		}
+        else {
+            // M for Misc. error
+            snprintf(msg, sizeof(msg), "  _M_ ");
+        }
+
+        // Add newline ever 16 addresses
+        if ((i + 1) % 16 == 0 && i > 0) {
+            snprintf(&msg[strlen(msg)], sizeof(msg) - strlen(msg), "\n");
+        }
+        strncat(response_output_buf, msg, remaining_space);
+	}
+    return 0;
+}
--- a/firmware/Core/Src/telecommands/i2c_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/i2c_telecommand_defs.c
@@ -4,14 +4,18 @@
 #include <string.h>
 #include "main.h"
 
+/// @brief Scans the I2C bus for devices. Prints out the addresses of devices found.
+/// @param args_str
+/// - Arg 0: I2C bus to scan (1-4)
+/// @return 0 if successful, 1 if error.
 uint8_t TCMDEXEC_scan_i2c_bus(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len) {
 
     I2C_HandleTypeDef* hi2c;
-    uint8_t bus_to_scan = atoi(args_str);
+    const uint8_t bus_to_scan = atoi(args_str);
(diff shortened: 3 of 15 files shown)


firmware/Core/Inc/telecommands/i2c_telecommand_defs.h
```diff
@@ -0,0 +1,7 @@
+#include <stdint.h>
+#include "telecommands/telecommand_definitions.h"
+#include "main.h"
```
DeflateAwning, line 3: This file needs an include guard (see other header files for examples).

Please include other header files from the `.c` files, and not from `.h` files.

firmware/Core/Inc/telecommands/i2c_telecommand_defs.h
```diff
@@ -0,0 +1,7 @@
...
+
+
+uint8_t TCMDEXEC_scan_i2c_device(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
```
DeflateAwning, line 6: Can we rename from `scan_i2c_device` to `scan_i2c_bus`? We're not scanning a single device, but rather scanning for any/all devices on a bus.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -0,0 +1,52 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+uint8_t TCMDEXEC_scan_i2c_device(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+
+    snprintf(response_output_buf, response_output_buf_len, "Starting I2C scan...\n");
+
+    I2C_HandleTypeDef* hi2c;
+    int bus_to_scan = atoi(args_str);
```
DeflateAwning, line 12: What happens if this fails? Also, please use `uint8_t` (or similar specific type) instead of `int`

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -0,0 +1,52 @@
...
+            break;
+        case 3: 
+            hi2c = &hi2c3;
+            break;
+        case 4:
+            hi2c = &hi2c4;
+            break;
+        default:
+            hi2c = NULL;
+            snprintf(response_output_buf, response_output_buf_len, "Enter a valid I2C bus!\n");
```
DeflateAwning, line 30: This over-writes the "Starting I2C scan..." message. Perhaps we want to remove the "Starting scan" message?

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -0,0 +1,52 @@
...
+            return 0;
+    }
+
+
+	char msg[5];
+	// Go through all possible i2c addresses
+	for (uint8_t i = 0; i < 128; i++) {
+
+		if (HAL_I2C_IsDeviceReady(hi2c, (uint16_t)(i<<1), 3, 5) == HAL_OK) {
```
DeflateAwning, line 39: What are the `3, 5` args? Looks like magic numbers, which are bad.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -0,0 +1,52 @@
...
+			// We got an ack
+			sprintf(msg, "%2x ", i);
+			snprintf(response_output_buf, response_output_buf_len, "%s", msg);
```
DeflateAwning, line 42: Doesn't this overwrite at the start?

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -233,5 +239,4 @@ uint8_t TCMDEXEC_available_telecommands(const char *args_str, TCMD_TelecommandCh
     snprintf(response_output_buf, response_output_buf_len, "%s", response);
 
     return 0;
-}
-
+}
```
DeflateAwning, line 242: Avoid unnecessary whitespace changes

DeflateAwning (changes requested):
Very solid start! Looks like there are some logical bugs and some "best practice" opportunities.

Please ensure you've tested code before you request a review. If you test this without any I2C devices connected, the output should represent that there are no I2C devices connected.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
DeflateAwning: Here, can we also please track the different return statuses from `HAL_I2C_IsDeviceReady` so that `HAL_TIMEOUT` is represented differently than `HAL_ERROR` ?

firmware/Core/Inc/telecommands/i2c_telecommand_defs.h
parkerg-13: Done.

firmware/Core/Inc/telecommands/i2c_telecommand_defs.h
parkerg-13: Done.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: If the conversion fails atoi returns 0 which is handled by the default case in the switch statement. Should this be handled explicitly before we hit the switch statement?

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: In reference to the 3, 5 args I can give them variable names. Are these numbers what we want? The 3 represents the number of trials to see if the device is ready and the 5 represents the timeout in milliseconds.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: I misunderstood the functioning of this way of writing to UART. Fix will be soon.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: Yes misunderstood how things were working this has been fixed.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: Error messages separated.

parkerg-13:
Still have to test with devices attached.

firmware/Core/Inc/rtos_tasks/rtos_tasks.h
```diff
@@ -5,6 +5,8 @@
 #include "main.h"
 #include "rtos_tasks/rtos_task_helpers.h"
 
+#define MAX_RESPONSE_BUFFER_LENGTH 1024
```
DeflateAwning, line 8: Requires namespace prefix (probably `TCMD_`)

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -0,0 +1,68 @@
...
+    HAL_StatusTypeDef i2c_device_status;
+
+	// Go through all possible i2c addresses
+	for (uint8_t i = 0; i < 128; i++) {
+
+        i2c_device_status = HAL_I2C_IsDeviceReady(hi2c, (uint16_t)(i<<1), 3, 5);
+
+		if (i2c_device_status == HAL_OK) {
+			// O for OK
+			sprintf(msg, "%2x O\n", i);
```
DeflateAwning, line 43: Please use `snprintf` for length overflow safety. No `snprintf`. `strcat` is also unsafe.

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -9,6 +9,7 @@
 #include "telecommands/flash_telecommand_defs.h"
 #include "telecommands/lfs_telecommand_defs.h"
 #include "telecommands/timekeeping_telecommand_defs.h"
+#include "i2c_telecommand_defs.h"
```
DeflateAwning, line 12: Please add folder prefix

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -0,0 +1,68 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include "main.h"
+
+uint8_t TCMDEXEC_scan_i2c_bus(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+
+    I2C_HandleTypeDef* hi2c;
+    uint8_t bus_to_scan = atoi(args_str);
```
DeflateAwning, line 11: What happens if `args_str` isn't valid? I don't think that's handled here currently

DeflateAwning (changes requested):
More comments. Good changes overall though

firmware/Core/Inc/rtos_tasks/rtos_tasks.h
parkerg-13: Done.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: Done.

firmware/Core/Src/telecommands/telecommand_definitions.c
parkerg-13: Done.

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
parkerg-13: What do you mean by not valid? If anything funky is passed to atoi then it will just return 0 and we will handle it as something going wrong in the switch statement.

DeflateAwning:
Changes look good. Can you please rebase to `main`?

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
DeflateAwning: For example, if `args_str = "this is not an integer"`, it's important we consider what happens. Because there's no "i2c 0", that works well enough here.

DeflateAwning (approved):
Great work, Parker G! 

I rebased to `main` so that this can be merged. I also added counts of each status to the end of the message. 

Feel free to add another commit in this branch to add your name to the list of contributors in the README! When you're ready, you can click "Squash and Merge" to merge this change into `main`.

DeflateAwning (dismissed):

DeflateAwning (approved):

PR #81, opened 2024-07-02
Added Enumeration for FLASH driver code to categorize errors

Saksham-P:
### Enum Created:

```enum FLASH_ERRORS {
    FLASH_ERR_OK                    = 0, // No error occurred
    FLASH_ERR_SPI_TRANSMIT_FAILED   = -3, // Error occurred while transmitting SPI signal
    FLASH_ERR_SPI_RECEIVE_FAILED    = -4, // Error occurred while receiving SPI signal
    FLASH_ERR_DEVICE_BUSY_TIMEOUT   = -6, // Took too long for the device to be in standby
    FLASH_ERR_UNKOWN                = -7, // Unknown error occurred (code reached where it shouldn't have been possible)
    FLASH_ERR_STATUS_REG_ERROR      = -8 // Error occurred which was indicated by one of the Status Register Bits.
};
```
The above are the values of the enum that are created to organize the type of errors we can expect in `.../littlefs/flash_driver.c`. The values have been cross-checked with LittleFS return enum values **only** in `lfs.h` file just in case they might conflict later on, so the integer values in the enum are unique to `.../littlefs/flash_driver.c`

### Changes Made:
Most return values have been replaced by the values of above enum, very few have been replaced by a variable, for e.g.:
```
const uint8_t read_status_result = FLASH_read_status_register(hspi, chip_number, status_reg_buffer);
    if (read_status_result != FLASH_ERR_OK) {
        return read_status_result;
    }
```
We are returning the variable since it contains the enum value of the error caused within `FLASH_read_status_register()`

### DIFF
--- a/firmware/Core/Inc/littlefs/flash_driver.h
+++ b/firmware/Core/Inc/littlefs/flash_driver.h
@@ -48,25 +48,25 @@ static const uint8_t FLASH_SR1_PROGRAMMING_ERROR_MASK = (1 << 6);
 static const uint8_t FLASH_SR1_ERASE_ERROR_MASK = (1 << 5);
 
 /*-----------------------------FLASH ERROR CODES-----------------------------*/
-enum FLASH_ERRORS {
+typedef enum {
     FLASH_ERR_OK                    = 0,    // No error occurred
     FLASH_ERR_SPI_TRANSMIT_FAILED   = -3,   // Error occurred while transmitting SPI signal
     FLASH_ERR_SPI_RECEIVE_FAILED    = -4,   // Error occurred while receiving SPI signal
     FLASH_ERR_DEVICE_BUSY_TIMEOUT   = -6,   // Took too long for the device to be in standby
-    FLASH_ERR_UNKNOWN                = -7,   // Unknown error occurred (code reached where it shouldn't have been possible)
+    FLASH_ERR_UNKNOWN               = -7,   // Unknown error occurred (code reached where it shouldn't have been possible)
     FLASH_ERR_STATUS_REG_ERROR      = -8    // Error occurred which was indicated by one of the Status Register Bits.
-};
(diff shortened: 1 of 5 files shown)


DeflateAwning:
Please rename pull request to describe the change (generally related to the issue that it closes).

DeflateAwning (changes requested):
Decent overall! Main issue is that the return types are `uint`, and you're returning an enum type. The return types should be marked as enums, and intermediate value variables should also be marked as enums.

Saksham-P:
I made the changes to make sure there are no type conflicts,  I was quite surprised that such a major conflict in type wouldn't give a build error.

I have also merged in the latest changes from 'main' branch and incorporated type changes into those. 

As for the whitespaces in `flash_driver.h`, if you like I can have those match the ones in `main` branch. 

Feel free to review the changes made, although I will probably review them myself and most likely wait till Thursday to merge into `main` even if no additional changes are needed.

DeflateAwning (changes requested):
Looks great overall!

Can we please add a new error enum values that's along the lines of `FLASH_ERR_SPI_HAL_TRANSMIT_TIMEOUT` and `FLASH_ERR_SPI_HAL_RECEIVE_TIMEOUT`, which looks at the HAL timeout? It's an important error that indicates a very low-level hardware comms failure. Sorry I didn't suggest it earlier.

Saksham-P:
Yea I could do that, while I do that, should I also incorporate other errors that can happen alongside such as `HAL_ERROR` and `HAL_BUSY`, or do you think we should be fine without those

DeflateAwning:
I think "HAL_TIMEOUT" and then "every other HAL error enum value" is probably adequate. The others generally come as a result of misconfiguration and not from run-time problems

DeflateAwning (approved):
Looks great, thanks!

PR #82, opened 2024-07-03
Implement `TCMD_extract_double_arg` function, `compare_floats` function

DeemDeem52:
Added functions for telecommand parsing, as well as number comparisons


DeflateAwning:
* Please do a clean build before opening a PR. Please fix build issues.
* Please follow the function documentation convention demonstrated in all other functions in the main branch, and [in the applicable docs section](https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/blob/main/docs/C_General_Guidelines.md#documentation)
* Please update the PR title to reflect the changes made, as it will become the commit message when you merge.

These look like super helpful changes! Thanks for splitting these out into a separate PR!

DeflateAwning (changes requested):
* Please do a clean build and the fix build issues.
* Please update the PR title to reflect the changes made, as it will become the commit message when you merge.

These look like super helpful changes! Thanks for splitting these out into a separate PR!

parker-research:
I fixed up the rebase situation.

The `TCMD_ascii_to_double` function isn't safe, unfortunately. It needs to validate the string is a valid float, and also be able to accept the string `"0.0"`; currently, it returns `0.0` when it fails.

DeemDeem52:
I think I've fixed the issue with TCMD_ascii_to_double. It now should return 0 if the string doesn't start with a valid double.

DeemDeem52:
Er... it should return 2, not 0. That was a typo.

DeflateAwning (changes requested):
I think we want `TCMD_ascii_to_double` to fail (return an error) if the input string doesn't contain a valid float from start to finish. In the case of a typo, we don't want to try to execute a command with garbled arguments. 

I'd love to see unit tests that show:

```c
double output_val;

// Nominal test.
TCMD_ascii_to_double("6.7", &output_val) == 0
// assert: output_val approx equals 6.7

// Nominal test (negative).
TCMD_ascii_to_double("-23.9", &output_val) == 0
// assert: output_val value

// Error: empty string
TCMD_ascii_to_double("", &output_val) != 0

// Error: just a space
TCMD_ascii_to_double(" ", &output_val) != 0

// Error: letters
TCMD_ascii_to_double("abc", &output_val) != 0

// Error: trailing letters (this is the example I was talking about above)
TCMD_ascii_to_double("23.7a", &output_val) != 0
```

DeflateAwning (changes requested):
To write the specification explictly, I mean that the entire argument (from the start of the string/from the character after the previous comma, to the end of the string/to the next comma) must be a valid float. No whitespace. No non-float characters. No valid float characters (period, negative sign) in invalid locations.

This shall be captured in TCMD_ascii_to_double, and in TCMD_extract_double_arg.

Here are some more test cases to capture this explicitly:

```c
TEST_ASSERT_TRUE(TCMD_ascii_to_double("a23.7a", 6, &output_val) != 0);
TEST_ASSERT_TRUE(TCMD_ascii_to_double("23a7", 4, &output_val) != 0);
TEST_ASSERT_TRUE(TCMD_ascii_to_double("23ar.7", 6, &output_val) != 0);
TEST_ASSERT_TRUE(TCMD_ascii_to_double(" 23.7", 5, &output_val) != 0); // has a space
TEST_ASSERT_TRUE(TCMD_ascii_to_double("23.7 ", 5, &output_val) != 0); // has a space
TEST_ASSERT_TRUE(TCMD_ascii_to_double("23.7.1", 6, &output_val) != 0); // two decimal places
TEST_ASSERT_TRUE(TCMD_ascii_to_double(".7", 2, &output_val) != 0); // can't have decimal at start
TEST_ASSERT_TRUE(TCMD_ascii_to_double("7.", 2, &output_val) != 0); // can't have decimal at end
```

Some of these do not currently pass.

A simpler solution is probably to do a single check like this at the start before calling atof:

```c
uint8_t number_of_decimal_points_found = 0;
for (i=0; i< str_len ; i++) { // main loop
// negative sign at start is allowed
if (i==0 && iter_char == '-') { continue; }

// decimal sign is allowed once, but not at the start/end
if (iter_char == '.' && number_of_decimal_points_found = 0 && (i != 0 && i!=str_len-1)) {
number_of_decimal_points_found += 1;
continue;
}

// if it's a number, continue
if (isdigit(iter_char)) { continue; }

// if we've reached the end of the valid conditions for this char, it's invalid
return 1; // invalid char
}

// now that we know they're valid chars, use atof
```

DeemDeem52:
Your changes worked perfectly! (And I'll have to keep in mind the `continue;` syntax for further use...)
I think it should be ready now.

firmware/Core/Inc/telecommands/telecommand_args_helpers.h
```diff
@@ -7,4 +7,11 @@ uint8_t TCMD_ascii_to_uint64(const char *str, uint32_t str_len, uint64_t *result
 uint8_t TCMD_extract_uint64_arg(const char *str, uint32_t str_len, uint8_t arg_index, uint64_t *result);
 uint8_t TCMD_extract_string_arg(const char *str, uint8_t arg_index, char *result, uint16_t result_max_len);
 uint8_t TCMD_extract_hex_array_arg(const char *args_str, uint8_t arg_index, uint8_t result_array[], uint16_t result_array_size, uint16_t *result_length);
+
+uint8_t TCMD_ascii_to_double(const char *str, uint32_t str_len, double *result);
```
DeflateAwning, line 11: Awesome work overall! Just a tiny issue here with duplicate function prototypes. 

Also, this would be a good opportunity for you to add your name to the list of contributors in the README :)

DeflateAwning (changes requested):
Sorry, just one tiny change.

Also, this would be a good opportunity for you to add your name to the list of contributors in the README :)

DeemDeem52:
Good catch! And thank you for all your help!

DeflateAwning (approved):
Awesome, happy to help! :) 

Feel free to squash and merge, then delete your branch, when you get a chance!

PR #88, opened 2024-07-06
Create flash/LFS timing benchmark telecommands

### LINKED ISSUE #57: LFS: Benchmark read/write performance, investigate opportunities to increase write performance
* Setup a macro to disable debugging UART prints in the critical path of LFS writes (i.e., in the FLASH and LFS drivers/implementations).
* Create a telecommand to benchmark the read/write performance by reading and writing large files. Output a report with the speeds.
* Find ways to optimize these speeds, if required (TBC).

parker-research:
* Create flash module and LFS timing benchmark telecommands
* Switch LFS to not use `malloc` via its macro
* Rename global `lfs`, `cfg`, and `file_cfg` variables

Closes #57 (benchmark LFS)

### DIFF
--- a/firmware/Core/Inc/config/static_config.h
+++ b/firmware/Core/Inc/config/static_config.h
@@ -4,11 +4,11 @@
 
 /// Whether to enable critical-path debug success logging in the FLASH drivers.
 /// Default: 0 (disabled)
-#define FLASH_ENABLE_UART_DEBUG_PRINT 1
+#define FLASH_ENABLE_UART_DEBUG_PRINT 0
 
 /// Whether to enable critical-path debug success logging in the LittleFS drivers.
 /// Default: 0 (disabled)
-#define LFS_ENABLE_UART_DEBUG_PRINT 1
+#define LFS_ENABLE_UART_DEBUG_PRINT 0
 
 
 // NOTE: All C files that use config variables from here should include a guard like the following:
--- a/firmware/Core/Inc/config/static_config.h
+++ b/firmware/Core/Inc/config/static_config.h
@@ -2,19 +2,14 @@
 #ifndef __INCLUDE_GUARD__STATIC_CONFIG_H__
 #define __INCLUDE_GUARD__STATIC_CONFIG_H__
 
-/// Whether to enable critical-path debug success logging in the FLASH drivers.
+#include <stdint.h>
+
+/// Whether to enable hot-path debug success logging in the FLASH drivers.
 /// Default: 0 (disabled)
-#define FLASH_ENABLE_UART_DEBUG_PRINT 0
+static const uint8_t FLASH_enable_hot_path_debug_logs = 0;
 
-/// Whether to enable critical-path debug success logging in the LittleFS drivers.
+/// Whether to enable hot-path debug success logging in the LittleFS drivers.
 /// Default: 0 (disabled)
-#define LFS_ENABLE_UART_DEBUG_PRINT 0
-
-
-// NOTE: All C files that use config variables from here should include a guard like the following:
-// #ifndef LFS_ENABLE_UART_DEBUG_PRINT
-//     #error "LFS_ENABLE_UART_DEBUG_PRINT not defined"
-// #endif
-
+static const uint8_t LFS_enable_hot_path_debug_logs = 0;
 
 #endif /* __INCLUDE_GUARD__STATIC_CONFIG_H__ */
--- a/firmware/Core/Inc/config/static_config.h
+++ b/firmware/Core/Inc/config/static_config.h
@@ -11,4 +11,10 @@
 #define LFS_ENABLE_UART_DEBUG_PRINT 1
 
 
+// NOTE: All C files that use config variables from here should include a guard like the following:
+// #ifndef LFS_ENABLE_UART_DEBUG_PRINT
+//     #error "LFS_ENABLE_UART_DEBUG_PRINT not defined"
+// #endif
+
+
 #endif /* __INCLUDE_GUARD__STATIC_CONFIG_H__ */
--- a/firmware/Core/Inc/config/static_config.h
+++ b/firmware/Core/Inc/config/static_config.h
@@ -1,0 +1,14 @@
+
+#ifndef __INCLUDE_GUARD__STATIC_CONFIG_H__
+#define __INCLUDE_GUARD__STATIC_CONFIG_H__
+
+/// Whether to enable critical-path debug success logging in the FLASH drivers.
+/// Default: 0 (disabled)
+#define FLASH_ENABLE_UART_DEBUG_PRINT 1
+
+/// Whether to enable critical-path debug success logging in the LittleFS drivers.
+/// Default: 0 (disabled)
+#define LFS_ENABLE_UART_DEBUG_PRINT 1
+
+
+#endif /* __INCLUDE_GUARD__STATIC_CONFIG_H__ */
--- a/firmware/Core/Src/littlefs/flash_driver.c
+++ b/firmware/Core/Src/littlefs/flash_driver.c
@@ -5,9 +5,6 @@
 #include "debug_tools/debug_uart.h"
 
 #include "config/static_config.h"
-#ifndef FLASH_ENABLE_UART_DEBUG_PRINT
-    #error "FLASH_ENABLE_UART_DEBUG_PRINT not defined"
-#endif
 
 /// Timeout duration for HAL_SPI_READ/WRITE operations.
 // Note: FLASH_read_data has sporadic timeouts at 5ms; 10ms is a safe bet.
@@ -153,11 +150,11 @@ uint8_t FLASH_write_enable(SPI_HandleTypeDef *hspi, uint8_t chip_number)
             return 2;
         }
 
-        #if FLASH_ENABLE_UART_DEBUG_PRINT
-        DEBUG_uart_print_str("DEBUG: status_reg = 0x");
-        DEBUG_uart_print_array_hex(status_reg_buffer, 1);
-        DEBUG_uart_print_str("\n");
-        #endif
+        if (FLASH_enable_hot_path_debug_logs) {
+            DEBUG_uart_print_str("DEBUG: status_reg = 0x");
+            DEBUG_uart_print_array_hex(status_reg_buffer, 1);
+            DEBUG_uart_print_str("\n");
+        }
     }
 
     // Should never be reached:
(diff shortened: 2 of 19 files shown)


parker-research:
@JohnathanBurchill Whenever you get a chance, let me know if you think this is reasonable merge into `main` (ideally with an Approval). 

I know there's still more work to do on benchmarking and trying to solve the performance issues, but I think this is a reasonable start on it at least.

We're also considering switching from NOR flash to NAND flash, which offers significantly faster erase times (2ms per block instead of 500ms per block), which would be an easy alternative to digging too deep into LFS's performance.

firmware/Core/Inc/config/static_config.h
```diff
@@ -0,0 +1,20 @@
+
+#ifndef __INCLUDE_GUARD__STATIC_CONFIG_H__
```
JohnathanBurchill, line 2: Why do we use leading underscores in the include guard? Aren't they reserved for the compiler implementation?

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -4,7 +4,15 @@
 #include "littlefs/flash_driver.h"
 #include "debug_tools/debug_uart.h"
 
+#include "config/static_config.h"
+#ifndef FLASH_ENABLE_UART_DEBUG_PRINT
+    #error "FLASH_ENABLE_UART_DEBUG_PRINT not defined"
```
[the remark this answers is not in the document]
JohnathanBurchill: Just read the comment explaining this above. The #error confused me, though, because a common idiom is that simply #define'ing toggles the compilation. It seems error prone and cumbersome to have to remember to include this guard. 

Suggest to make this clear in the coding style doc.

firmware/Core/Src/littlefs/littlefs_benchmark.c
```diff
@@ -0,0 +1,153 @@
...
+    uint8_t write_buffer[write_chunk_size];
+    for (uint32_t i = 0; i < write_chunk_size; i++) {
+        const uint8_t b = (i + 42) % 256;
+        write_buffer[i] = b;
+    }
+
+    // Open file for writing
+    const uint32_t write_open_start_time = HAL_GetTick();
+    lfs_file_t file;
+    const int8_t open_result = lfs_file_opencfg(&LFS_filesystem, &file, file_name, LFS_O_WRONLY | LFS_O_CREAT | LFS_O_TRUNC, &LFS_file_cfg);
```
JohnathanBurchill, line 32: Why not `lfs_file_open` instead of `lfs_file_opencfg`?

firmware/Core/Src/littlefs/littlefs_benchmark.c
```diff
@@ -0,0 +1,153 @@
...
+
+/// @brief Benchmarks the write/read operations on the LittleFS file system.
+/// @details This function will write test data to a static filename, read it back, and verify the read data.
+/// @param write_chunk_size Number of bytes to write in each chunk.
+/// @param write_chunk_count Number of chunks to write.
+/// @param response_str 
+/// @param response_str_len 
+/// @return 0 on success. >0 if there was an error.
+uint8_t LFS_benchmark_write_read(uint16_t write_chunk_size, uint16_t write_chunk_count, char* response_str, uint16_t response_str_len) {
+    const char file_name[] = "benchmark_test.txt";
```
JohnathanBurchill, line 17: Suggest `"benchmark_test.txt"` be `#defined` in a `filenames.h` header to be included by any translation unit needing file access, or make it an argument to the function. Formalizing this should prevent name clashes across subsystems.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -19,12 +25,13 @@ uint8_t LFS_is_lfs_mounted = 0;
 uint8_t LFS_read_buffer[FLASH_CHIP_PAGE_SIZE_BYTES];
 uint8_t LFS_prog_buffer[FLASH_CHIP_PAGE_SIZE_BYTES];
 uint8_t LFS_lookahead_buf[16];
```
JohnathanBurchill, line 27: Maybe `#define` also the lookahead buffer size.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -274,37 +293,44 @@ int8_t LFS_write_file(char *file_name, uint8_t *write_buffer, uint32_t write_buf
...
 	lfs_file_t file;
-	const int8_t open_result = lfs_file_open(&lfs, &file, file_name, LFS_O_RDONLY);
+	const int8_t open_result = lfs_file_opencfg(&LFS_filesystem, &file, file_name, LFS_O_RDONLY, &LFS_file_cfg);
 	if (open_result < 0)
 	{
 		DEBUG_uart_print_str("Error opening file to read\n");
 		return open_result;
 	}
 	
+    #if LFS_ENABLE_UART_DEBUG_PRINT
```
JohnathanBurchill, line 306: Same question about indentation.

firmware/Core/Src/main.c
```diff
@@ -82,7 +83,7 @@ const osThreadAttr_t TASK_DEBUG_print_heartbeat_Attributes = {
 osThreadId_t TASK_handle_uart_telecommands_Handle;
 const osThreadAttr_t TASK_handle_uart_telecommands_Attributes = {
   .name = "TASK_handle_uart_telecommands",
-  // Size 2048 doesn't work with LFS settings, but 8192 does
+  // Size 2048 doesn't work with LFS settings, but 8192 does.
   // TODO: confirm stack size
   .stack_size = 8192,
```
JohnathanBurchill, line 88: Suggest to `#define` the stack size.

JohnathanBurchill (commented):
Overall looks good - was not able to test the LFS benchmark - possible issue with my dev kit setup? Should be looked into, but not a stopper for merging this into `main`.

The LFS GitHub README has this note:

"[SPIFFS](https://github.com/pellepl/spiffs) - Another excellent embedded filesystem for NOR flash. As a more traditional logging filesystem with full static wear-leveling, SPIFFS will likely outperform littlefs on small memories such as the internal flash on microcontrollers."

Considered already?

JohnathanBurchill:
> Considered already?

Just looked at it - not suited to large memories and not for real-time ops. Though I like the idea of a flat directory structure.

JohnathanBurchill:
> We're also considering switching from NOR flash to NAND flash, which offers significantly faster erase times (2ms per block instead of 500ms per block), which would be an easy alternative to digging too deep into LFS's performance.

Not sure what to make of this. Would be good to discuss.

firmware/Core/Src/littlefs/flash_driver.c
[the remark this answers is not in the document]
parker-research: Hmm that's fair. The fear was that if you forget to include the header file in the place you're using the configuration variable, then it takes on a value silently. The idea was that, by requiring a check that the macro is present (regardless of its value), it prevents that issue. 

It's an annoying quirk of C that `#if` assumes a default value instead of erroring when the macro is unavailable.

Thinking about it more, `static const` variables are probably the best approach in achieving the safety I was going for. I'll switch to that.

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -145,9 +153,11 @@ uint8_t FLASH_write_enable(SPI_HandleTypeDef *hspi, uint8_t chip_number)
             return 2;
         }
 
+        #if FLASH_ENABLE_UART_DEBUG_PRINT
```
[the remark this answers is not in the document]
parker-research: The idea was to target hot-path locations first. It's not a huge deal if errors take 2ms longer with a debug print. I will rename the macro to clarify that though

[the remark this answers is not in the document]
parker-research: Yes, although maybe I should conform better to the C style haha

firmware/Core/Inc/config/static_config.h
parker-research: You're right, totally forgot about that!

The intention was to put them at the "top" of a sorted list so they're not in the way in Intellisense suggestions. This indeed breaks the rule though.

Whether it's worth it to refactor and re-communicate that change may be open for debate though; the chance of the compiler creating this exact format of macro is probably pretty low (albeit non-zero).

firmware/Core/Src/littlefs/littlefs_benchmark.c
parker-research: `lfs_file_opencfg` uses a statically-allocated "file buffer" (in the `LFS_file_cfg` struct), which is required to prevent LFS from calling `malloc`. That change is a non-benchmarking change in this PR (which maybe should have been separated out into a separate PR, ideally).

firmware/Core/Src/main.c
parker-research: Stack size is only referenced in one location. Unnecessary abstraction in my opinion. Each FreeRTOS task has its own stack size, so defining it right with the task seems reasonable.

parker-research:
We did consider SPIFFS in the informal filesystem "trade study". I believe the conclusions were largely:

* Its wear leveling mechanism was much weaker
* It didn't have support for handling "bad blocks", which seemed like a reasonable feature, especially in a space setting. 
* Its recommended max filesystem size was lower than we were planning
* LittleFS was more self-contained, and had a more clear "getting started" path on what needed to be implemented to get an initial integration going. The "getting started" path for SPIFFS was less clear. 
* The idea was generally just that both are suitable, but that LFS is probably more suitable, and so we'd aim to use it first if we could get it working, and would try SPIFFS if we encountered significant issues.

firmware/Core/Src/littlefs/littlefs_benchmark.c
```diff
@@ -0,0 +1,153 @@
...
+    const uint32_t open_end_time = HAL_GetTick();
+    snprintf(
+        &response_str[strlen(response_str)],
+        response_str_len - strlen(response_str),
+        "Write open: %lu ms\n", open_end_time - write_open_start_time);
+
+    // Write
+    const uint32_t write_send_start_time = HAL_GetTick();
+    for (uint32_t chunk_num = 0; chunk_num < write_chunk_count; chunk_num++) {
+        const int8_t write_result = lfs_file_write(&LFS_filesystem, &file, write_buffer, write_chunk_size);
```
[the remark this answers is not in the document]
parker-research: Hmm now that I think about it, I could see why that would be nice, especially as you're reviewing that .c file.

As far as I can tell though, the standard practice is to put the extern declarations in the associated header file. It ensures consistency, decreases the number of places that require an update if there's a rename/type change, and re-inforces the idea that the extern variables are intended to be accessed alongside the other contents of that header file.

parker-research:
Implemented suggestions in one commit. Thanks for the review; great feedback and comments overall! :)

JohnathanBurchill (approved):
Seems fine.

PR #107, opened 2024-07-15
Telecommand scheduling: suffix tag parser, executor thread, agenda/schedule

### LINKED ISSUE #15: Command scheduling: schedule a command for execution at a specific timestamp

parker-research:
**Main Change:** Schedule command exeuction in the future with: `CTS1+hello_world()@tsexec=25000!`

### Features:

* Created suffix tag parser (for parsing `@tsexec=uint64` values), with unit tests
* Implemented telecommand executor thread (`TASK_execute_telecommands`), and moved execution from the UART receive thread to this executor thread
* Developed telecommand agenda/schedule (`TCMD_agenda` array)
* Isolated parsing from execution steps (parsing in umbilical UART receive thread still)

### Other Updates:

* Added warning in TIM_set_current_unix_epoch_time_ms if a resync timestamp is before the previous resync timestamp, as that case will create confusing log messages (but should never happen in reality). 

Closes #15


firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -0,0 +1,175 @@
...
+        DEBUG_uart_print_str(" !!!!!! ERROR !!!!!!");
+    }
+    DEBUG_uart_print_str(") ========\n");
+    DEBUG_uart_print_str(response_output_buf);
+    DEBUG_uart_print_str("\n==========================\n");
+
+    return tcmd_result;
+}
+
+/// @brief Executes a telecommand immediately.
```
JohnathanBurchill, line 150: Could clarify the difference between this function and the function above. Both use the same `@brief` summary.

JohnathanBurchill (changes requested):
Looks great! Suggest to open new issues for

- providing a telecommand to list the current telecommand agenda, or to list the next N commands to be executed; Or immediately (or periodically) writing the agenda to an onboard file, which can be read at startup, in case of power cycle; Or keep track of everything on ground.
- providing a telecommand to remove a telecommand from the agenda, perhaps by telecommand ID, or by its location in the agenda, or for deleting all commands from the agenda, pending development of a strategy for maintaining the agenda across power cycles.
- Revise or start documentation about mission operations and procedures, such as how to schedule commands. 

Overall, this is a solid contribution.

parker-research:
Thanks for the review! Fixed that comment, and opened 3 new issues to capture those agenda-related operations. Great comments! 

I'll advise that persisting the agenda to the filesystem and reading it from the filesystem (now Issue #110), while nice to have, is probably a task that would be better to leave until all other mission requirements are satisfied.

JohnathanBurchill (approved):
Thanks, looks good.

PR #118, opened 2024-07-22
Updated Flash driver and LittleFS helper to use NAND memory module (MT29F1G Series)

Saksham-P:
## Update
Replaced all driver code to encompass the NAND memory module commands from the [datasheet](https://www.farnell.com/datasheets/3151163.pdf). The code has not yet been tested, and should be tested first using the function `FLASH_is_reachable()`

For writing to the module, we first write to the cache register, and then run another command which reads the cache register and writes to the main memory array. 
For reading to the module, we first send a command which reads the data to a cache register and send another command which outputs the data from the cache register. 

## Comments
I am not sure if Reading and Writing will work right away since in the documentation for reading from cache (Page 18) and documentation for writing to the cache (Page 29, 30) doesn't specify if we should be checking Status Register to see and wait for data to load to and from cache registers, it only states that read/write operations from cache register will end when CS# is driven to HIGH.

### DIFF
--- a/firmware/Core/Inc/littlefs/flash_driver.h
+++ b/firmware/Core/Inc/littlefs/flash_driver.h
@@ -8,44 +8,50 @@
 
 #include "littlefs/lfs.h"
 
-
 /*----------------------------- CONFIG VARIABLES ----------------------------- */
-
-// number of CS pins available
+// Number of CS pins available
 #define FLASH_NUMBER_OF_FLASH_DEVICES 8 // TODO: update to 8, or 10 with FRAM maybe
-#define FLASH_CHIP_SIZE_BYTES 67108864  // 64MiB // TODO: update
+
+// Total size of a singular Memory Module in bytes
+#define FLASH_CHIP_SIZE_BYTES 134217728  // 128MiB // TODO: update
+
+/*-------------------------------FLASH FEATURES-------------------------------*/
+// Features that can be accessed using Get Feature command
+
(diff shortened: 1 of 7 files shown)


DeflateAwning (commented):
Overall, this looks really solid. Thanks for working on this! I'm exicted to start testing it very soon, and will do a more thorough review once we've found any trial-and-error bugs.

DeflateAwning:
@KaleF07 Any luck testing/debugging this? No worries either way

KaleF07:
> @KaleF07 Any luck testing/debugging this? No worries either way

Still working through, will make some progress over the next couple of days.

DeflateAwning:
Just rebased this onto main. If using GitHub Destop, you'll need to delete your local branch and then re-pull from `origin` (GitHub Online). If using git CLI, you'll need to checkout this branch (`git checkout saksham-NAND-flash-driver`), do a `git fetch -a` then `git reset --hard origin/saksham-NAND-flash-driver`

DeflateAwning:
Potentially-helpful hint: `flash_read_hex` command doesn't seem to be working. Reading at address=0 gets the same bytes as reading at address=1. Of course, reading from addr=1 should yield the bytes from addr=0, but starting from the second byte.

NuclearTea:
@KaleF07 @Saksham-P Because #177 Was merged, I think this can be closed?

DeflateAwning:
I supported closing this and deleting the branch to avoid confusion. Will wait for @Saksham-P's confirmation though!

Saksham-P:
I agree, I will close the PR, and also delete the branch

Issue #116, opened 2024-07-20
Logging channel/sink for in-memory log failsafe


JohnathanBurchill:
(suggested by @DeflateAwning) "Add a channel which writes to an in-memory string circular buffer (maybe 4096 bytes long) which would store the most recent ~50-100 logs. Could be super helpful in case the filesystem is experiencing a failure, and the radio downlinks aren't received/logged with 100% reliability."

DeflateAwning:
I think this was fixed in #119, and that there's nothing left to do with it. Would you mind confirming, @JohnathanBurchill? Feel free to close it if this is the case.

JohnathanBurchill:
Fixed in #119.

PR #119, opened 2024-07-23
In-memory LOG storage for most recent messages

JohnathanBurchill:
Addresses issue #116.

The in-memory log table serves as statically-allocated log message storage that allows concurrent and nested calls to LOG_message() and serves as a limited backup for the filesystem logs. As such, this is not considered a traditional log sink.

Logs are stored in a circular buffer.

Adds telecommands for reporting the latest and the N-latest log messages from in-memory log table.

### DIFF
--- a/firmware/Core/Src/log/log.c
+++ b/firmware/Core/Src/log/log.c
@@ -19,7 +19,7 @@
 #define LOG_SYSTEM_NAME_MAX_LENGTH 20
 // Messages up to 256 characters
 #define LOG_FORMATTED_MESSAGE_MAX_LENGTH 256
-// Includes prefix, with cushion for delimiters and newline
+// Includes prefix, with cushion for delimiters, newline, and null terminator
 #define LOG_FULL_MESSAGE_MAX_LENGTH ( LOG_FORMATTED_MESSAGE_MAX_LENGTH + LOG_TIMESTAMP_MAX_LENGTH + LOG_SINK_NAME_MAX_LENGTH + LOG_SYSTEM_NAME_MAX_LENGTH + 1 )
 
 typedef struct {
@@ -189,7 +189,7 @@ void LOG_message(LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_
                 case LOG_SINK_UNKNOWN:
                 default:
                     // Should not reach this except by memory corruption 
-                    LOG_to_umbilical_uart("Error: unkown log sink\n");
+                    LOG_to_umbilical_uart("Error: unknown log sink\n");
                     break;
             }
         }
@@ -446,7 +446,7 @@ const char *LOG_get_memory_table_full_message_at_index(uint8_t index)
 }
 
 /// @brief Get the most recent log message text 
-/// @return pointer to the full text of the most recent lost message
+/// @return pointer to the full text of the most recent message
 /// (statically allocated)
 const char *LOG_get_most_recent_log_message_text(void)
 {
--- a/firmware/Core/Src/log/log.c
+++ b/firmware/Core/Src/log/log.c
@@ -22,10 +22,6 @@
 // Includes prefix, with cushion for delimiters and newline
 #define LOG_FULL_MESSAGE_MAX_LENGTH ( LOG_FORMATTED_MESSAGE_MAX_LENGTH + LOG_TIMESTAMP_MAX_LENGTH + LOG_SINK_NAME_MAX_LENGTH + LOG_SYSTEM_NAME_MAX_LENGTH + 1 )
 
-static char LOG_timestamp_string[LOG_TIMESTAMP_MAX_LENGTH] = {0};
-static char LOG_formatted_log_message[LOG_FORMATTED_MESSAGE_MAX_LENGTH] = {0};
-static char LOG_full_log_message[LOG_FULL_MESSAGE_MAX_LENGTH] = {0};
-
 typedef struct {
     LOG_sink_enum_t sink;
     const char name[LOG_SINK_NAME_MAX_LENGTH];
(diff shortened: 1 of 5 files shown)


firmware/Core/Src/log/log.c
```diff
@@ -392,3 +422,34 @@ const char* LOG_get_severity_name(LOG_severity_enum_t severity)
...
+/// specified index
+/// @param index requested in-memory log message
+/// @return pointer to the full text of the log message
+const char *LOG_get_memory_table_full_message_at_index(uint8_t index)
+{
+    return LOG_memory_table[index].full_message;
+}
+
+/// @brief Get the most recent log message text 
+/// @return pointer to the full text of the most recent lost message
```
DeflateAwning, line 449: Typo: "lost message"

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -195,3 +197,80 @@ uint8_t TCMDEXEC_log_set_system_debugging_messages_state(const char *args_str, T
...
+
+    // Report the entries in the log table, which is a circular buffer
+    uint8_t current_index = LOG_get_memory_table_index_of_most_recent_log_entry();
+    int16_t start_index = ((int16_t)current_index + 1 - (int16_t)requested_number_of_entries); 
+    if (start_index < 0) {
+        start_index += max_entries;
+    }
+    uint8_t index = (uint8_t)start_index;
+    void (*transmit_function)(const char *);
+    switch (tcmd_channel) {
```
DeflateAwning, line 251: Thoughts on switching this to just call the function directly (instead of function pointer)? Feels like the function pointer doesn't add a ton.

If-statement would be fine, or switch statement is good too I think

firmware/Core/Src/log/log.c
```diff
@@ -146,19 +176,19 @@ void LOG_message(LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_
...
+                    LOG_to_umbilical_uart(current_log_entry->full_message);
                     break;
-                // FIXME: handle case LOG_SINK_MEMORY
+                case LOG_SINK_NONE:
+                case LOG_SINK_UNKNOWN:
                 default:
-                    // Recursion not allowed; use direct calls
-                    // Should not reach this anyway
+                    // Should not reach this except by memory corruption 
                     LOG_to_umbilical_uart("Error: unkown log sink\n");
```
DeflateAwning, line 192: Typo in unknown

DeflateAwning (approved):
Looks solid to me! 

The solution I had in mind looked more like a circular buffer with delimiters between log messages, but this seems solid! No need to complicate and change things. 

A few tiny semantic changes if you want, but looks very good!

firmware/Core/Src/log/log.c
JohnathanBurchill: Fixed

firmware/Core/Src/telecommands/log_telecommand_defs.c
JohnathanBurchill: The rationale was to pull that logic out of the loop, which could be up to 128 iterations. Speed is not critical here, but it seems cleaner and slightly more efficient.

firmware/Core/Src/log/log.c
JohnathanBurchill: fixed

firmware/Core/Src/telecommands/log_telecommand_defs.c
DeflateAwning: Oh fair; I don't think the time cost of 128 if-statement evaluations is too significant. This sort of optimization is mostly only important in very time-critical code, or where there are operations which take many many compute instructions (e.g., memory reads/writes of arrays).

Will leave the final choice up to you though!

JohnathanBurchill: Will leave it as it is then. Please approve the PR again.

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -195,3 +197,80 @@ uint8_t TCMDEXEC_log_set_system_debugging_messages_state(const char *args_str, T
...
+        case TCMD_TelecommandChannel_DEBUG_UART:
+            transmit_function = LOG_to_umbilical_uart;
+            break;
+        case TCMD_TelecommandChannel_RADIO1:
+            transmit_function = LOG_to_umbilical_uart;
```
DeflateAwning, line 256: I think this should be `LOG_to_uhf_radio` or similar. I think at the very least, there should be a `FIXME` comment here so we don't miss it.

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -195,3 +197,80 @@ uint8_t TCMDEXEC_log_set_system_debugging_messages_state(const char *args_str, T
...
+            break;
+        default:
+            // Channel is unknown, log error to the filesystem and return
+            LOG_message(LOG_SYSTEM_TELECOMMAND, LOG_SEVERITY_ERROR, LOG_SINK_FILE, "TCMDEXEC_log_report_latest_message(): Unknown telecommand channel");
+            return 1;
+            break;
+    }
+
+    // Report the messages
+    for (uint8_t i = 0; i < requested_number_of_entries; i++) {
```
DeflateAwning, line 266: OPTIONAL: What about this:
```c
for (uint8_t response_num = 0; response_num < requested_number_of_entries; response_num++) {
const uint8_t log_index = (current_index + response_num) % max_entries;
}
```

Incrementing a counter within a for-loop is a bit of an anti-pattern, imo. The modulo operator was designed for wrap-arounds.

May need to make the variables slightly wider, as I think the uint8+uint8 part may overflow.

DeflateAwning (dismissed):
Just noticed two more things scrolling through quickly. Change is approved if you disagree with them, but just wanted to comment anyway! :)

firmware/Core/Src/telecommands/log_telecommand_defs.c
JohnathanBurchill: Nice catch on that transmit function. I like the modulo arithmetic approach. 

Ready for approval.

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -195,3 +197,76 @@ uint8_t TCMDEXEC_log_set_system_debugging_messages_state(const char *args_str, T
...
+        default:
+            // Channel is unknown, log error to the filesystem and return
+            LOG_message(LOG_SYSTEM_TELECOMMAND, LOG_SEVERITY_ERROR, LOG_SINK_FILE, "TCMDEXEC_log_report_latest_message(): Unknown telecommand channel");
+            return 1;
+            break;
+    }
+
+    // Report the messages
+    for (uint8_t response_number = 0; response_number < requested_number_of_entries; response_number++) {
+        const uint16_t log_index = (start_index + response_number) % max_entries;
```
DeflateAwning, line 266: It might be fine, but I have a strange fear that the `start_index + response_number` intermediate value may be cast to a `uint8` (as response_number is a uint8). I can't find a source that confirms this one way or another. For safety (unless we can confirm it's fine), would you mind making `response_number` a uint16 or int16?

I've approved the PR, and you can merge it if you're highly confident this intermediate value can't/won't overflow or wrap around.

DeflateAwning (approved):

firmware/Core/Src/telecommands/log_telecommand_defs.c
JohnathanBurchill: That was an oversight. Would rather use 16-bit numbers to be safe. Updated, and ready to approve.

DeflateAwning (approved):
Looks great! Thanks for bearing with my nitpicking :)

PR #120, opened 2024-07-23
Telecommand extract int64 function

DeemDeem52:
New telecommand function for signed integers

### DIFF
--- a/firmware/Core/Src/telecommands/telecommand_args_helpers.c
+++ b/firmware/Core/Src/telecommands/telecommand_args_helpers.c
@@ -128,7 +128,7 @@ uint8_t TCMD_ascii_to_int64(const char *str, uint32_t str_len, int64_t *result) 
 /// @param result Pointer to the result
 /// @return 0 if successful, 1 if the string is empty, 2 if the string does not contain enough arguments
 ///        3 if the argument is not an integer, 4 for other error
-uint8_t TCMD_extract_int64_arg(const uint8_t *str, uint32_t str_len, uint8_t arg_index, int64_t *result) {
+uint8_t TCMD_extract_int64_arg(const char *str, uint32_t str_len, uint8_t arg_index, int64_t *result) {
     if (str_len == 0) {
         return 1;
     }
--- a/firmware/Core/Src/telecommands/telecommand_args_helpers.c
+++ b/firmware/Core/Src/telecommands/telecommand_args_helpers.c
@@ -83,6 +83,82 @@ uint8_t TCMD_extract_uint64_arg(const char *str, uint32_t str_len, uint8_t arg_i
     return 0;
 }
 
+/// @brief Extracts the longest substring of integer characters, starting from the beginning of the
+///     string, to a maximum length or until the first non-integer character is found.
+/// @param str Input string, starting with an integer
+/// @param str_len Max length of the input string
+/// @param result Pointer to the result
+/// @return 0 if successful, 1 if the string is empty, 2 if the string does not start with an integer
+uint8_t TCMD_ascii_to_int64(const uint8_t *str, uint32_t str_len, int64_t *result) {
+    if (str_len == 0) {
+        return 1;
+    }
+    uint64_t temp_result = 0;
+    uint32_t i = 0;
+    for (; i < str_len; i++) {
+        if (str[i] == '-' && i == 0) {
+            continue;
+        }
+        else if (str[i] < '0' || str[i] > '9') {
+            break;
+        }
+        temp_result = temp_result * 10 + (str[i] - '0');
+    }
+
+    if (str[0] == '-') {
+        temp_result *= -1;
+    }
+
+    if (i == 0) {
+        return 2;
+    }
+
+    *result = temp_result;
+    return 0;
+}
+
+/// @brief Extracts the nth comma-separated argument from the input string, assuming it's an int64
+/// @param str Input string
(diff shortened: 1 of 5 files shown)


DeemDeem52:
The unit test for this function passes, and there are no warnings or errors when clean building.

firmware/Core/Src/telecommands/telecommand_args_helpers.c
```diff
@@ -83,6 +83,86 @@ uint8_t TCMD_extract_uint64_arg(const char *str, uint32_t str_len, uint8_t arg_i
...
+        return 1;
+    }
+    uint64_t temp_result = 0;
+    uint32_t i = 0;
+    for (; i < str_len; i++) {
+        if (str[i] == '-' && i == 0) {
+            continue;
+        }
+        else if (str[i] < '0' || str[i] > '9') {
+            break;
```
parker-research, line 103: I think this should be an error return, not a break.

firmware/Core/Src/telecommands/telecommand_args_helpers.c
```diff
@@ -83,6 +83,86 @@ uint8_t TCMD_extract_uint64_arg(const char *str, uint32_t str_len, uint8_t arg_i
     return 0;
 }
 
+/// @brief Extracts the longest substring of integer characters, starting from the beginning of the
```
parker-research, line 86: String parsing functions on satellite missions should not do "best attempt at parsing the longest valid response". They should know definitively if the string is fully valid or not. 

If we sent the `turn_off_forever(189000O000`) telecommand, where the arg is the timestamp to turn off forever, do we want the parser to extract `189000` (the longest numerical string from the start) because there's a typo with a letter `O` in the string? Or do we want it to say "I think you made a typo; try again"?

firmware/Core/Src/telecommands/telecommand_args_helpers.c
```diff
@@ -83,6 +83,86 @@ uint8_t TCMD_extract_uint64_arg(const char *str, uint32_t str_len, uint8_t arg_i
...
+///     string, to a maximum length or until the first non-integer character is found.
+/// @param str Input string, starting with an integer
+/// @param str_len Max length of the input string
+/// @param result Pointer to the result
+/// @return 0 if successful, 1 if the string is empty, 2 if the string does not entirely encapsulate an integer
+uint8_t TCMD_ascii_to_int64(const char *str, uint32_t str_len, int64_t *result) {
+    if (str_len == 0) {
+        return 1;
+    }
+    uint64_t temp_result = 0;
```
parker-research, line 96: Temp result is unsigned? You multiply by negative 1 later.

firmware/Core/Src/telecommands/telecommand_args_helpers.c
```diff
@@ -83,6 +83,86 @@ uint8_t TCMD_extract_uint64_arg(const char *str, uint32_t str_len, uint8_t arg_i
...
+}
+
+/// @brief Extracts the nth comma-separated argument from the input string, assuming it's an int64
+/// @param str Input string
+/// @param str_len Length of the input string
+/// @param arg_index Index of the argument to extract (0-based)
+/// @param result Pointer to the result
+/// @return 0 if successful, 1 if the string is empty, 2 if the string does not contain enough arguments
+///        3 if the argument is not an integer, 4 for other error
+uint8_t TCMD_extract_int64_arg(const char *str, uint32_t str_len, uint8_t arg_index, int64_t *result) {
```
parker-research, line 131: Would love to see a unit test or two for this function!

parker-research (changes requested):
Good start!

firmware/Core/Src/telecommands/telecommand_args_helpers.c
DeemDeem52: There's no need for that; it's dealt with already immediately afterwards.

firmware/Core/Src/telecommands/telecommand_args_helpers.c
DeemDeem52: Oops! I did write the function that way, but I forgot to fix the comment. Good catch -- thanks!

firmware/Core/Src/telecommands/telecommand_args_helpers.c
DeemDeem52: Thanks!

firmware/Core/Src/telecommands/telecommand_args_helpers.c
DeemDeem52: Done! Thank you!

firmware/Core/Src/telecommands/telecommand_args_helpers.c
parker-research: Is there a case where this "break" is hit and the result isn't to return? I don't think so.

Returning immediately avoids errors. This is convoluted, imo.

parker-research (commented):
Looks solid! Thanks for all the changes and unit tests. Just one comment I want to discuss more.

DeemDeem52:
I see what you mean. Fixed, now!

DeflateAwning (approved):
I just pushed a small change. Give it a check over and make sure you're happy with it, and then please Squash and Merge it!

parker-research (approved):

DeemDeem52:
Thank you very much Parker! Squashed and merged.

PR #121, opened 2024-07-24
Resolved FIXME's for the TCMD_ascii_to_uint64() helper function

KaleF07:
Added error handling for FIXME's present in the TCMD_ascii_to_uint64() helper function, added unit tests for possible errors.

### DIFF
--- a/firmware/Core/Src/telecommands/telecommand_args_helpers.c
+++ b/firmware/Core/Src/telecommands/telecommand_args_helpers.c
@@ -13,14 +13,22 @@
 /// @param str Input string, starting with an integer
 /// @param str_len Max length of the input string
 /// @param result Pointer to the result
-/// @return 0 if successful, 1 if the string is empty, 2 if the string does not start with an integer
+/// @return 0 if successful, 1 if the string is empty, 2 if the string does not start with an integer,
+///         3 if the size of result doesn't match str_len, 4 for if the string is too long
 uint8_t TCMD_ascii_to_uint64(const char *str, uint32_t str_len, uint64_t *result) {
     // FIXME: return error if the string is too long/number is too large
     // FIXME: return error if the number doesn't occupy the whole string (e.g., "123abc" with str_len=6 should error)
     // TODO: write unit tests for this function
     // TODO: consider removing the str_len parameter and using strlen(str) instead (requires refactor in caller)
     if (str_len == 0) {
+        *result = 0;
         return 1;
+    }
+
+    // Return error if size of integer is larger than the total length of the result variable (20 for uint64_t)
+    if (str_len > 20) {
+        *result = 0;
+        return 4;
     }
 
     uint64_t temp_result = 0;
@@ -34,7 +42,12 @@ uint8_t TCMD_ascii_to_uint64(const char *str, uint32_t str_len, uint64_t *result
     }
 
     if (i == 0) {
+        // Error if the string does not start with an integer
+        *result = 0;
         return 2;
+    } else if ( i != str_len) {
+        // Error if result length deosn't match str_len
+        return 3;
     }
 
     *result = temp_result;
--- a/firmware/Core/Src/telecommands/telecommand_args_helpers.c
+++ b/firmware/Core/Src/telecommands/telecommand_args_helpers.c
@@ -13,27 +13,24 @@
 /// @param str Input string, starting with an integer
 /// @param str_len Max length of the input string
 /// @param result Pointer to the result
-/// @return 0 if successful, 1 if the string is empty, 2 if the input string doesn't match str_len, 3 for if the string is too long,
-///         4 if the result size doesn't match str_len (includes non-integer characters)
+/// @return 0 if successful, 1 if the string is empty, 2 if the string size is longer than 19 characters,
+///         3 if the string contains non-integer characters.
+/// @note Result is not written to if an error occurs. The max digit length that can be store in the result
+/// is 19. If the input string is longer than 19 characters, an error will be returned.
 uint8_t TCMD_ascii_to_uint64(const char *str, uint32_t str_len, uint64_t *result) {
+    // FIXME: return error if the string is too long/number is too large
+    // FIXME: return error if the number doesn't occupy the whole string (e.g., "123abc" with str_len=6 should error)
+    // TODO: write unit tests for this function
     // TODO: consider removing the str_len parameter and using strlen(str) instead (requires refactor in caller)
 
     // Error: the input string is empty
-    if (str_len == 0 || strlen(str) == 0) {
-        *result = 0;
+    if (str_len == 0) {
         return 1;
     }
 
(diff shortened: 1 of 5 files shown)


firmware/Core/Src/telecommands/telecommand_args_helpers.c
```diff
@@ -13,16 +13,29 @@
...
-    if (str_len == 0) {
+
+    // Error: the input string is empty
+    if (str_len == 0 || strlen(str) == 0) {
+        *result = 0;
         return 1;
     }
 
+    // Error: the input string doesn't match size of str_len
+    if(strlen(str) != str_len) {
```
DeflateAwning, line 28: The way this function is setup to work (and is used) is actually such that you can pass it a pointer to a point within a string (not necessarily at the start), and only allow it to operate on a given length of the substring by setting the `str_len` to less than `strlen(str)`. This is a technique borrowed from Rust ("slicing" a string), which is a little uncommon in C.

This check, as-is, will cause this function to not work at all, I'm pretty sure. That's what my TODO comment was intended to indicate with the "requires refactor in caller" comment.

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
```diff
@@ -1,6 +1,35 @@
...
 #include "unit_tests/unit_test_helpers.h"
 
+uint8_t TEST_EXEC__TCMD_ascii_to_uint64() {
+    uint64_t result;
+
+    // Success: Nominal test
+    TEST_ASSERT(TCMD_ascii_to_uint64("123", 3, &result) == 0);
+    TEST_ASSERT(result == 123);
+
+    // Error: Empty string
```
DeflateAwning, line 11: The maximum value of a `uint64_t` is `18446744073709551615` (herein `UINT64_MAX`). Test cases for `UINT64_MAX`, `UINT64_MAX+1`, and `UINT64_MAX-1` would be valuable.

I believe this function currently overflows if you pass it `UINT64_MAX-1` (or any other 20-digit number larger than this max). 

I believe setting the max "digit length" to 19 characters would be appropriate (even though it does prevent passing the UINT64_MAX as a max value). A comment in the docstring explaining the limitation would mitigate this restriction.

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
```diff
@@ -1,6 +1,35 @@
...
+    TEST_ASSERT(TCMD_ascii_to_uint64("", 0, &result) == 1);
+    TEST_ASSERT(result == 0);
+
+    // Error: String size doesn't match str_len
+    TEST_ASSERT(TCMD_ascii_to_uint64("123456", 10, &result) == 2);
+    TEST_ASSERT(result == 0);
```
DeflateAwning, line 17: Please remove these `result == 0` on error; the result should be unchanged on error.

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
```diff
@@ -1,6 +1,35 @@
...
+
+    // Error: String is too long
+    TEST_ASSERT(TCMD_ascii_to_uint64("123456789123456789123", 21, &result) == 3)
+    TEST_ASSERT(result == 0);
+
+    // Error: String doesn't start with an integer
```
DeflateAwning, line 23: Comment on line 23 doesn't describe the code that follows

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
```diff
@@ -1,6 +1,35 @@
...
+    TEST_ASSERT(TCMD_ascii_to_uint64("abc123", 6, &result) == 4)
+    TEST_ASSERT(result == 0);
+
+    // Error: Result length doesn't match str_len
```
DeflateAwning, line 27: Result length? Result is a number, and doesn't have a "length" really

DeflateAwning (changes requested):
Solid start. Just a couple issues. Make sure you test it with the demo telecommand which parses several int arguments. 

It would also be valuable to write unit tests for the `TCMD_extract_uint64_arg()` function, as I believe it may be broken as-is.

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
KaleF07: I updated so that the max digit length is 19 characters and added a note describing the restriction in the docstring.

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
KaleF07: Updated comment to accurately match the error

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
KaleF07: Makes sense, I switched to say "digit length" instead

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
```diff
@@ -1,6 +1,55 @@
...
+}
+
+uint8_t TEST_EXEC__TCMD_extract_uint64_arg() {
+    uint64_t result;
+    char str[] = "1,2,3,4,a,6,7,8,9";
+    // Success: Nominal Test
+    TEST_ASSERT(TCMD_extract_uint64_arg(str, sizeof(str), 0, &result) == 0);
+    TEST_ASSERT(result == 1);
+
+    char str2[0];
```
KaleF07, line 37: I know that making an array of size 0 is illegal, however is this alright for the purpose of unit tests?

DeflateAwning: I'd strongly advise against it. An array of size 1, with the argument currently passed at `sizeof(str2)` set explicity to 0, would be safer.

Best as I can tell, `char str2[0];` is undefined behaviour by the C standard, and relies on the given C compiler to implement well.

DeflateAwning (changes requested):
One test case change, then will approve

DeflateAwning (changes requested):
Solid. Almost ready to approve. 

Can you please just update the `char str2[0];` line?

Or, if I have a chance later, I'll fix it and get ready to merge it.

DeflateAwning:
Also, feel free to add your name to the README as a contributor!

firmware/Core/Src/unit_tests/test_telecommand_arg_helpers.c
KaleF07: Fixed!

DeflateAwning (approved):
Sweet! Squash and merge. Thanks!

~~Don't forget to eventually add your name to the README contributor list if you want!~~

PR #124, opened 2024-07-25
Telecommand to extract FreeRTOS stack information

christopherkinyua:
This ticket addresses creating a telecommand for obtaining the task stack information using FreeRTOS APIs.

The output is structured as a JSON for easy viewing. The information I extracted is the: Task Name, State, Priority Number, and the StackHighWaterMark, which is the amount of stack that remained unused when the task stack was at its greatest (deepest) value.

PR #127, opened 2024-07-27
Telecommand to extract FreeRTOS stack information

christopherkinyua:
This PR addresses creating a telecommand for obtaining the task stack information using FreeRTOS APIs.

I have addressed issues and comments based on an earlier iteration of a previous PR  [#124 ](https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/pull/124)

I have added a new source and header file for the freeRTOS telecommand. I have changed how I was printing to use `DEBUG_uart_print_str`

### DIFF
--- a/firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
+++ b/firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
@@ -3,9 +3,13 @@
 #define __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
 
 #include <stdint.h>
+#include <FreeRTOS.h>
+#include <task.h>
 #include "telecommands/telecommand_types.h"
 
-uint8_t TCMDEXEC_retrieve_freertos_metadata(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+const char* freertos_current_state_enum_to_str(eTaskState state);
+
+uint8_t TCMDEXEC_freertos_get_metadata(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len);
 
 #endif // __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
--- a/firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
+++ b/firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
@@ -3,13 +3,8 @@
 #define __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
 
 #include <stdint.h>
-#include <FreeRTOS.h>
-#include <task.h>
-#include "telecommands/telecommand_types.h"
 
-const char* freertos_current_state_enum_to_str(eTaskState state);
-
-uint8_t TCMDEXEC_freertos_get_metadata(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+uint8_t TCMDEXEC_freetos_list_tasks_jsonl(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len);
 
 #endif // __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
--- a/firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
+++ b/firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
@@ -1,0 +1,11 @@
+
+#ifndef __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
+#define __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
+
+#include <stdint.h>
+#include "telecommands/telecommand_types.h"
+
+uint8_t TCMDEXEC_retrieve_freertos_metadata(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len);
+
+#endif // __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
--- a/firmware/Core/Src/telecommands/freertos_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/freertos_telecommand_defs.c
@@ -24,47 +24,51 @@
 /// @param response_output_buf The buffer to write the response to
 /// @param response_output_buf_len The maximum length of the response_output_buf (its size)
 /// @return 0 if successful, >0 if an error occurred (but hello_world can't return an error)
+const char* freertos_current_state_enum_to_str(eTaskState state) {
+    switch (state) {
+        case eRunning:    return "RUN";
+        case eReady:      return "RDY";
+        case eBlocked:    return "BLK";
+        case eSuspended:  return "SUSP";
+        case eDeleted:    return "DEL";
+        default:          return "UNKN";
+    }
+}
 
-uint8_t TCMDEXEC_retrieve_freertos_metadata(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+uint8_t TCMDEXEC_freertos_get_metadata(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len) {
                             
-    TaskStatus_t *pxTaskStatusArray;
-    UBaseType_t uxArraySize, x;
-    unsigned long ulTotalRunTime;
+    TaskStatus_t *px_task_status_array;
+    UBaseType_t ux_array_size, x;
+    unsigned long ul_total_run_time;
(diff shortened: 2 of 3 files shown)


firmware/Core/Src/telecommands/freertos_telecommand_defs.c
```diff
@@ -0,0 +1,70 @@
...
+                               x < uxArraySize - 1 ? "," : "");
+            if (len > 0) {
+                DEBUG_uart_print_str(message_buffer);
+            }
+        }
+
+        DEBUG_uart_print_str("]");
+        vPortFree(pxTaskStatusArray);
+        
+        } else {
```
DeflateAwning, line 64: Please invert the if-statement above, so that all error conditions return early, and then the "happy path" code is one indent level less-deep at the bottom of the function.

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
```diff
@@ -0,0 +1,70 @@
...
+#include <task.h>
+#include <portmacro.h>
+
+
+/// @brief Telecommand that return metadata regarding Tasks from FreeRTOS
+/// @param args_str No arguments expected
+/// @param tcmd_channel The channel on which the telecommand was received, and on which the response should be sent
+/// @param response_output_buf The buffer to write the response to
+/// @param response_output_buf_len The maximum length of the response_output_buf (its size)
+/// @return 0 if successful, >0 if an error occurred (but hello_world can't return an error)
```
DeflateAwning, line 26: Check this comment. Please remove empty lines between `///` comment block and function definition.

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
```diff
@@ -0,0 +1,70 @@
...
+
+    if (pxTaskStatusArray != NULL) {
+
+        uxArraySize = uxTaskGetSystemState(pxTaskStatusArray, uxArraySize, &ulTotalRunTime);
+        DEBUG_uart_print_str("[");
+
+        for (x = 0; x < uxArraySize; x++) {
+            char message_buffer[256];
+            int len = snprintf(message_buffer, sizeof(message_buffer),
+                               "{\"TaskName\":\"%s\",\"State\":\"%s\",\"Priority\":%lu,\"Stack\":%u}%s",
```
DeflateAwning, line 46: Can you please change the keys to be snake case (e.g., "task_name", etc.)

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -249,6 +250,17 @@ const TCMD_TelecommandDefinition_t TCMD_telecommand_definitions[] = {
 
    // ****************** END SECTION: log_telecommand_defs ******************
 
+   // ****************** SECTION: freertos_telecommand_defs ******************
+
+   {
+        .tcmd_name = "retrieve_freertos_metadata",
```
DeflateAwning, line 256: Just to keep with the namespace "theme" of "most important info first", would you mind renaming to `freertos_get_metadata`? I'd normally think you name makes more sense, but this way the commands sort a bit nicer. Update function name and string here.

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
```diff
@@ -0,0 +1,70 @@
...
+                               pxTaskStatusArray[x].pcTaskName,
+                               pxTaskStatusArray[x].eCurrentState == eRunning ? "RUN" :
```
DeflateAwning, line 48: Can you please make a `const char * freertos_current_state_enum_to_str(enum)` (or similar) function which returns a pointer to one of these static strings? It's allowed in C, and will make this part a little cleaner.

There's an example of this type of function in the logging stuff, if you wanted to check what an enum-to-string function can look like.

DeflateAwning (changes requested):
Very awesome work! Just a few more comments. Will approve after all these tweaks are made

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
christopherkinyua: Sure thing. Just addressed this

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
christopherkinyua: Done

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
christopherkinyua: Done

firmware/Core/Src/telecommands/telecommand_definitions.c
christopherkinyua: Done

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
christopherkinyua: Done, basically wrote a function to take the TaskState type as an argument and did the checking in there using a switch.

christopherkinyua:
Should I also delete the old branch where I was initially working on?

DeflateAwning:
Yeah, you're good to delete the old branch! When you merge this one, you'll be good to delete this branch too.

firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
```diff
@@ -0,0 +1,15 @@
+
+#ifndef __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
+#define __INCLUDE_GUARD__FREERTOS_TELECOMMAND_DEFINITIONS_H
+
+#include <stdint.h>
+#include <FreeRTOS.h>
+#include <task.h>
```
DeflateAwning, line 7: Please include only the bare minimum necessary header files in our `.h` files, and include the rest in the relevant `.c` files. Here, you should only need `stdint.h`, I think.

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
```diff
@@ -0,0 +1,74 @@
...
+        return 1;
+    }
+
+    ux_array_size = uxTaskGetSystemState(px_task_status_array, ux_array_size, &ul_total_run_time);
+    DEBUG_uart_print_str("[");
+
+    for (x = 0; x < ux_array_size; x++) {
+        char message_buffer[256];
+        int len = snprintf(message_buffer, sizeof(message_buffer),
+                            "{\"TaskName\":\"%s\",\"State\":\"%s\",\"Priority\":%lu,\"Stack\":%u}%s",
```
DeflateAwning, line 59: Please change the keys here to be snake case. Looks like it just got overlooked.

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
```diff
@@ -0,0 +1,74 @@
...
+
+    if (px_task_status_array == NULL) {
+        DEBUG_uart_print_str("Failed to allocate memory for task status array.");
+        return 1;
+    }
+
+    ux_array_size = uxTaskGetSystemState(px_task_status_array, ux_array_size, &ul_total_run_time);
+    DEBUG_uart_print_str("[");
+
+    for (x = 0; x < ux_array_size; x++) {
```
DeflateAwning, line 56: Can you please move the type for `x` to be here so that `x` is only scoped to the for loop. Standard style.

DeflateAwning (changes requested):
Awesome changes, just a couple more, sorry

firmware/Core/Src/telecommands/freertos_telecommand_defs.c
christopherkinyua: Ahh sure thing. I only changed the variables at that time. Thanks for the catch.

firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
christopherkinyua: I added the `#include <FreeRTOS.h>` and  `#include <task.h>` because in the `const char* freertos_current_state_enum_to_str(eTaskState state)`  function, I pass in an argument of type `eTaskState` which becomes unknown when I remove those header files. 

Is there another way around this? Unless I create my own custom type and make sure it kinda matches what FreeRTOS has. Idk, feels like a long work around. Let me know what you think

DeflateAwning: Because that function is only used in the one `.c` file, you don't even need to include it in a `.h` file at all

DeflateAwning:
I just tacked on a commit which fixed the ordering of some comments, fixed my comments about header organization, and removed the use of malloc. 

Please take a look and let me know if you have any questions or comments. If all looks good, please "Squash and Merge".

DeflateAwning (approved):

firmware/Core/Inc/telecommands/freertos_telecommand_defs.h
christopherkinyua: Ahh I see, so i wouldn't even need to add it to the header file since it isn't being reused else where.

christopherkinyua:
Thanks for the PR. Its interesting how you went about solving the ticket. Ill do my best next time to ensure this goes smoother than it did right now.

Issue #109, opened 2024-07-15
Telecommand Agenda: delete elements from the agenda (`agenda_delete_by_tssent`, `agenda_delete_all`)


parker-research:
Create a telecommand to remove a telecommand from the agenda.

## Telecommand 1: `agenda_delete_by_tssent`
Delete an entry from the agenda by its unique `timestamp_sent` field.

Argument: `tssent` value of the command to delete.

## Telecommand 2: `agenda_delete_all`
Delete all entries from the agenda. No arguments.

## Notes
* Agenda entries are "deleted" by marking them as invalid using the `TCMD_agenda_is_valid` boolean array.

PR #154, opened 2024-08-08
Telecommands to delete elements from the agenda 

christopherkinyua:
I created two telecommands to:
- delete all agendas
- delete an agenda using a timestamp sent arg 

Comments
- I was not sure where to put the telecommands and I kept them in `telecommands_definitions.c`. Let me know the best practice for this. I did not want to put it into the `telecommands_executor.c` file.
- I wanted to know if there is a way to test the telecommand. I only tested it by ensuring I was able to see and call the telecommand. I was not able to validate I got rid of the telecommand agenda. Is there a way to test if I deleted an agenda?


DeflateAwning:
Please move telecommands to a different file (named something like `agenda_telecommands.c/h`). Each category of telecommand should have its own file.

To test, you can schedule a telecommand for execution in the near future (e.g., `CTS1+hello_world()@tssent=1723170620007@tsexec=1000!`), and then try cancelling it by its tssent number.

Implements #109.

DeflateAwning (changes requested):
Looks super solid! Just want to move these into a separate file, then can be approved! Good work :)

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -173,3 +173,33 @@ uint8_t TCMD_execute_telecommand_in_agenda(const uint16_t tcmd_agenda_slot_num,
...
+            TCMD_agenda_is_valid[slot_num] = 0;
+        }
+    }
+    return 0;
+}
+
+/// @brief Deletes a telecommand from the agenda by its `timestamp_sent` field.
+/// @param tssent The `timestamp_sent` value of the telecommand to delete.
+/// @return 0 on success, 1 if the telecommand was not found.
+uint8_t TCMD_agenda_delete_by_tssent(uint64_t *tssent) {
```
DeflateAwning, line 191: Just noticed it, but can you please change the `tssent` arg to be a regular into instead of a pointer.

firmware/Core/Src/telecommands/agenda_telecommands.c
```diff
@@ -0,0 +1,64 @@
...
+    uint8_t parse_result = TCMD_extract_uint64_arg(args_str, strlen(args_str), 0, &tssent);
+
+    //Checking if the parsing was right
+    if (parse_result > 0) {
+        snprintf(response_output_buf, response_output_buf_len, "Error parsing timestamp sent: Err=%d", parse_result);
+        return 1;
+    }
+
+    // Pass the tssent into the function that handles the delete from the stack
+    uint8_t result = TCMD_agenda_delete_by_tssent(&tssent);
```
DeflateAwning, line 46: Make const.

firmware/Core/Src/telecommands/agenda_telecommands.c
```diff
@@ -0,0 +1,64 @@
...
+/// @param response_output_buf The buffer to write the response to
+/// @param response_output_buf_len The maximum length of the response_output_buf (its size)
+/// @return 0 on success, >0 on error
+uint8_t TCMDEXEC_agenda_delete_by_tssent(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+    
+    uint64_t tssent = 0;
+
+    // Parse the arg string passed into a uint64_t for the timestamp sent
+    uint8_t parse_result = TCMD_extract_uint64_arg(args_str, strlen(args_str), 0, &tssent);
```
DeflateAwning, line 37: Make const.

firmware/Core/Src/telecommands/agenda_telecommands.c
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    // Pass the tssent into the function that handles the delete from the stack
+    uint8_t result = TCMD_agenda_delete_by_tssent(&tssent);
+
+    if (result == 1)
+    {
+        DEBUG_uart_print_str("Telecommand with tssent: ");
+        DEBUG_uart_print_uint64(tssent);
+        DEBUG_uart_print_str(" not found in agenda.\n");
```
DeflateAwning, line 52: Please move return here, and then unnest the `if (result == 0)` part

christopherkinyua:
I've implemented the changes you suggested but wanted your insight based on the testing.

In the photo below, I have highlighted:
- The scheduled telecommand
- Attempt 1 of trying to delete the telecommand agenda
- Attempt 2 of trying to delete the telecommand agenda

My test process was:
1) Schedule a telecommand for 5000ms from the current epoch time which was 1723180260
2) Used the `agenda_delete_by_tssent` telecommand to delete the agenda using the tssent: 1723180260
3) Used the `agenda_delete_by_tssent` telecommand to delete the agenda using the tssent: 1723180260 added with the 5000 offset which is: 1723185260

The result was that they both failed and from my understanding, the first one should passed and be deleted from the agenda. I was just curious to see what your thoughts are and if something is missing in my process.

agenda_redline

DeflateAwning (changes requested):
Sorry I didn't catch these comments in the review earlier. Couple of tiny changes

DeflateAwning:
Regarding your testing procedure, `@tssent=5000` isn't an offset; it's an absolute unix timestamp in ms. As you can see, the command is executed immediately because the system uptime (which matches the current unix timestamp, until the `set_unix_timestamp_ms` telecommand (or whatever it's called) is run.

firmware/Core/Src/telecommands/agenda_telecommands.c
christopherkinyua: Done

firmware/Core/Src/telecommands/agenda_telecommands.c
christopherkinyua: Done

firmware/Core/Src/telecommands/agenda_telecommands.c
christopherkinyua: Done

firmware/Core/Src/telecommands/telecommand_executor.c
christopherkinyua: Done, I removed the pointer tag from the function and the function definition and passed the `const uint8_t tssent` variable into the function instead of a pointer.

Is there a reason I should avoid using a pointer for that instant?

christopherkinyua:
I'm trying to make sense of this. There are two tags that were passed: `tssent` which is the time the telecommand was sent and `tsexec` which I think is the time the telecommand should execute after time has passed.

So during the testing, I was doing, I was passing the current epoch time into `tssent` and adding the 5000 to the `tsexec`. The way I made the function is that it would use the `tssent` that I passed to be checked and then delete the agenda.

Unless the epoch time on the dev board is different, is there something I'm not getting? Like in my test process or the values I'm passing in?

firmware/Core/Src/telecommands/telecommand_executor.c
DeflateAwning: Using a pointer is unnecessary, allows for the introduction of accidental errors. Additionally, it prevents the value from being inadvertently modified from within the function.

DeflateAwning:
The `tssent` argument is currently mostly just used as a unique ID of each telecommand sent, currently. 

The board assumes it's 1970-01-01T00:00:00Z when it powers on. You may find it helpful to use the `CTS1+set_system_time(time_ms)!` command to set the time to the real current time for realistic testing. 

When the board powers on, its uptime is 0, and its unix epoch time is 0. Thus, the uptime (echoed in the heartbeat messages) and the unix timestamp are equal until the unix timestamp is set via the `set_system_time` telecommand, or by the GPS.

firmware/Core/Src/telecommands/telecommand_executor.c
christopherkinyua: Thanks that makes sense

christopherkinyua:
Ahh I sort of get it. So the `tssent` is like an ID as you said and the `tsexec` is the amount of time in ms that should pass then to run the telecommand which is relative to the system uptime.

Thanks, I was able to run a successful test of the telecommand as seen below. Im gonna wait for your approval then I can squash and merge.

agenda

DeflateAwning:
> tsexec is the amount of time in ms that should pass then to run the telecommand

No. `tsexec` is an absolute unix timestamp (number of miliseconds since 1970-01-01T00:00:00) at which to execute.

christopherkinyua:
Ahh yes, that's right. I failed to explain it like that.

DeflateAwning:
To facilitate testing and review, I'm making a couple changes/additions:

* `agenda_telecommands.c` and `agenda_telecommand_defs.h` should have the same name stem (probably `agenda_telecommand_defs` for consistency). I'm renaming `agenda_telecommands.c` to `agenda_telecommand_defs.c`.
* `TCMD_agenda_delete_all` cannot fail, and its failure code is not checked, so it is changed to a void function. 
* Added a log message showing which command and tssent/tsexec value is being executed from the agenda. 
* Improved logging within the two functions you created (now shows the name of the telecommand which is deleted from the agenda). 

Please review these changes and let me know if you have any questions or comments!

Thank you again for working on this!

DeflateAwning (approved):
After you've checked through the changes I tacked on, please squash-and-merge!

christopherkinyua:
It looks good and it makes sense to me. I see you used the `LOG_message()` function. I'll start using it now too. Ill squash and merge.

DeflateAwning:
Thanks a lot!

There'll be announcement about the new `LOG_message()` function soon! Just wanted to start using it a bit, as it shows the timestamp (which was useful to debug when the commands were/weren't being executed here).

Issue #25, opened 2024-05-18
Implement Over-the-Air (OTA) firmware upgrade


DeflateAwning:

JohnathanBurchill:
@DeflateAwning asked me to comment about a suggestion to look at xboot: 

https://github.com/alexforencich/xboot?tab=readme-ov-file

It's for ATMEL chips, but might provide some ideas or could be ported to STM32?

JohnathanBurchill:
The MPI uses it, and I've found it to be reliable. An old version of the MPI had two PCBs each with a microcontroller, and we could do OTA updates on both boards, with one as a passthrough to the other. That was a little less reliable though.

JohnathanBurchill:
Can receive memory pages and write to a file via telecommand. Will wait for Issue #47 to be pulled to main to complete uploading to the MPI.

NuclearTea:
After investigating this for a bit, there are a couple of things that we have noticed.

1. The current firmware is approx. ~280 Kilobytes in size (this is the release build).
The firmware version will be a huge challenge (especially if we are planning to store multiple copies or if we plan to keep one copy but replace the entire file every time we would like to update it.

There were a couple of ideas that arose from this that @robotoshi, @2BlackCats and myself were discussing.

Toshi mentioned that we can use a lossless compression algorithm to uplink less data and from his initial testing, he has reported ~18% compression (I believe this means that the new size is 82% of the original, please confirm this @robotoshi).

Additionally, Alex suggested that maybe we should not uplink the entire file and rather, use some sort of paging, or diffing algorithm on the ground station and only update the portions of the firmware that would change. This seems like a good idea however, implementing this is quite tricky and I don't think that it's something that we should hand implement (maybe there is a library that can do this for us?). Additionally, we would need ensure that we have an existing copy of the firmware in the memory module.

2. Uplinking the data is a much more difficult (for testing) then actually loading, swapping, and running the new firmware.
My current plan is the following:

- Use the cube programmer to upload the new firmware into the second internal flash bank.
- Load the firmware from the second internal flash bank into FLASH memory
- clear the second internal flash bank
- Run tcmd that takes file that we just created and load it into the second flash bank
- Run tcmd to switch to second flash bank

The above steps will test the following: **Given that there is a file in the FLASH memory, load it into the internal flash, and then switch.**

What needs to be figured out: 
- How to uplink the data in a reasonable manner
[...]

Issue #24, opened 2024-05-18
Implement bootloader to restore "golden copy" of firmware from FRAM storage


DeflateAwning:

NuclearTea:
After researching and discussing with Toshi, we will utilize the dual banks in the flash memory, where the first bank will be the running Application while the second bank will be the Golden Copy. In order continue working on this, we will need to be able to upload a golden copy which will require its own telecommand and write the flash banks. Since this will be useful for #18, I will create another issue and continue working on that.

PR #174, opened 2024-09-03
Create API which can read, write, and erase the internal flash memory

NuclearTea:
- These telecommands are not meant to be used in production, they are purely for convenience when testing #24 and #25


firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -363,6 +364,30 @@ const TCMD_TelecommandDefinition_t TCMD_telecommand_definitions[] = {
         .readiness_level = TCMD_READINESS_LEVEL_FOR_OPERATION
     },
     // ****************** END: MPI_telecommand_definitions ********************
+    // ****************** START SECTION: flash_bank_telecommand_defs ******************
+
+    {
+        .tcmd_name = "flash_bank_read",
+        .tcmd_func = TCMDEXEC_Flash_Bank_Read,
+        .number_of_args = 1,
+        .readiness_level = TCMD_READINESS_LEVEL_FLIGHT_TESTING,
```
DeflateAwning, line 373: If these telecommands aren't meant to be used in flight, can you please set an appropriate level for them here?

NuclearTea: Moved to ground usage only. Thanks for the catch

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+        {
+            // If the program operation fails, break the loop
+            break;
+        }
+    }
+
+    // Lock the Flash to disable the flash control register access
+    HAL_FLASH_Lock();
+
+    return status;
```
DeflateAwning, line 57: Do not return a HAL_StatusTypeDef as a uint8_t.

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Reads data from the flash memory.
+ *
+ * This function takes in an address, a pointer to a buffer to store the read data, and the length of the data to be read.
+ * It checks if the address is within the valid range of the flash memory and returns an error if it's not.
+ * If the address is valid, it reads the data from the flash memory and stores it in the provided buffer.
+ *
+ * @param address The starting address in the flash memory where the data will be read from.
+ * @param buffer A pointer to the buffer where the read data will be stored.
```
DeflateAwning, line 68: Explicitly state that this array must be `length` bytes long, if that is the case.

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    // Lock the Flash to disable the flash control register access
+    HAL_FLASH_Lock();
+
+    return status;
+}
+
+/**
+ * Reads data from the flash memory.
```
DeflateAwning, line 61: The docstring should say explicitly that it only operates on BANK 2. 

Proposal: Instead of specifying the address, does it make more sense to specify the memory region (as an enum) and the byte offset from the start of that region?

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+    if (address < Flash_Partitions_FLASH_BANK2 || (address + length) > (FLASH_BANK2_END))
+    {
+        return 1; // Return error if address is out of range
+    }
+
+    for (uint32_t i = 0; i < length - 1; i++)
+    {
+        buffer[i] = *(uint8_t *)(address + i);
+    }
+    buffer[length] = '\0';
```
DeflateAwning, line 85: This doesn't make sense. Why is this array being null-terminated? Docstring indicates nothing of this behaviour, and it is non-idiomatic.

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @return 0 if the erase operation is successful, page error otherwise.
+ */
+uint32_t Flash_Bank_Erase(uint16_t start_page_erase, uint16_t number_of_pages_to_erase)
+{
+    __HAL_FLASH_CLEAR_FLAG(FLASH_FLAG_ALL_ERRORS);
+
+    HAL_FLASH_Unlock();
+
+    FLASH_EraseInitTypeDef EraseInitStruct =
```
DeflateAwning, line 108: Const? Or mutible?

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+        {
+            .TypeErase = FLASH_TYPEERASE_PAGES,
+            .Banks = FLASH_BANK_2,
+            .Page = start_page_erase,
+            .NbPages = number_of_pages_to_erase};
+
+    uint32_t PageError = 0;
+    const HAL_StatusTypeDef erase_status = HAL_FLASHEx_Erase(&EraseInitStruct, &PageError);
+    HAL_FLASH_Lock();
+    if (erase_status == HAL_OK)
```
DeflateAwning, line 118: The final return should be success. Non-final returns should catch error cases.

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+ * Erases a specified number of pages from the flash bank.
+ *
+ * This function takes in the starting page to erase and the number of pages to erase.
+ * It clears any existing flash errors, unlocks the flash, and then erases the specified pages.
+ * If the erase operation is successful, it returns 0. Otherwise, it returns the page error.
+ *
+ * @param start_page_erase The starting page to erase.
+ * @param number_of_pages_to_erase The number of pages to erase.
+ *
+ * @return 0 if the erase operation is successful, page error otherwise.
```
DeflateAwning, line 100: What is "page error"?

firmware/Core/Src/telecommands/flash_bank_telecommands_defs.c
```diff
@@ -0,0 +1,85 @@
...
+#include <string.h>
+
+/// @brief Write data to the internal flash bank starting from address 0x08000000
+/// @param args_str
+/// - Arg 0: The data to write as a string
+/// @note This telecommand is only for testing purposes, it is purposfully not fully fleshed out
+/// as there is no intention on using this. Update as needed
+/// @note Currently, this is only writes all data given as a string.
+/// Thus, it is difficult to write integers directly
+uint8_t TCMDEXEC_flash_bank_write(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel, char *response_output_buf, uint16_t response_output_buf_len)
```
DeflateAwning, line 16: Strongly recommend renaming this to `internal_flash_write_hex`. 

Strongly recommend using the argument as hex instead of as a c-string. Look at the other hex function to see how to do it.

Doesn't this function need an address to write at as an argument?

firmware/Core/Src/telecommands/flash_bank_telecommands_defs.c
```diff
@@ -0,0 +1,85 @@
...
+        snprintf(response_output_buf, response_output_buf_len, "Error: %lu", res);
+        return 1;
+    }
+    return 0;
+}
+
+/// @brief Read data from the internal flash bank starting from address 0x08000000
+/// @param args_str
+/// - Arg 0: The number of bytes to read as a uint64_t
+uint8_t TCMDEXEC_flash_bank_read(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel, char *response_output_buf, uint16_t response_output_buf_len)
```
DeflateAwning, line 30: Doesn't this function need an address or an offset to read from? 

What format does it display the value in? I recommend hex.

firmware/STM32L4R5ZITx_FLASH.ld
```diff
@@ -64,7 +64,9 @@ MEMORY
 RAM (xrw)      : ORIGIN = 0x20000000, LENGTH = 640K
 RAM2 (xrw)      : ORIGIN = 0x10000000, LENGTH = 64K
 RAM3 (xrw)      : ORIGIN = 0x20040000, LENGTH = 384K
-FLASH (rx)      : ORIGIN = 0x8000000, LENGTH = 2048K
+FLASH (rx)      : ORIGIN = 0x8000000, LENGTH = 1024K
```
DeflateAwning, line 67: Nice work figuring this out! 

The IOC code generation overwrites this file. I wonder if there's a way to configure this in the IOC file? Otherwise, it must be added as a `docs/Non-Critical_Notes/Deployment_Final_Checks.md` item to ensure this file has remained not-regenerated.

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    // Clear all FLASH flags before starting the operation
+    __HAL_FLASH_CLEAR_FLAG(FLASH_FLAG_ALL_ERRORS);
+
+    // Write data to flash
+    for (uint32_t currentAddress = address; currentAddress < endAddress; currentAddress += 8)
+    {
+        // currentAddress - address is the number of bytes already written
+        // const uint64_t double_word = *(uint64_t *)(data + (currentAddress - address));
```
DeflateAwning, line 41: Avoid commented code.

firmware/Core/Inc/flash_bank/flash_bank.h
```diff
@@ -0,0 +1,23 @@
+#ifndef __INCLUDE_GUARD__FLASH_BANK_H__
```
DeflateAwning, line 1: Recommend renaming everything "flash bank"-related (files, functions, etc.) to "stm32_internal_flash" probably. The external flash (4x flash modules) is more of a "flash bank" than the internal flash imo; the name doesn't inherently differentiate.

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+ * @param address The starting address in the flash memory where the data will be read from.
+ * @param buffer A pointer to the buffer where the read data will be stored.
+ * @param length The length of the data to be read.
+ *
+ * @return 0 if the read operation is successful, 1 if the address is out of range.
+ */
+uint8_t Flash_Bank_Read(uint32_t address, uint8_t *buffer, uint32_t length)
+{
+    // Check if the address is within the valid range of the flash memory
+    if (address < Flash_Partitions_FLASH_BANK2 || (address + length) > (FLASH_BANK2_END))
```
DeflateAwning, line 76: Good safety check! :)

DeflateAwning (changes requested):
This is awesome! Great work on this! 

If you make these recommended changes (esp. the telecommand ones, related to using hex format and having an address/offset argument), you'll be setup to use these commands to push the actual firmware golden copy/push the new copies of the firmware.

DeflateAwning:
Also, please rebase :)

firmware/Core/Inc/flash_bank/flash_bank.h
```diff
@@ -0,0 +1,23 @@
...
+/// @note look in the STM32L4R5ZITx_FLASH.ld file
+/// to see the address of each partition, update this as needed
+typedef enum
+{
+    Flash_Partitions_RAM = 0x20000000,
+    Flash_Partitions_RAM_2 = 0x10000000,
+    Flash_Partitions_RAM_3 = 0x20040000,
+    Flash_Partitions_FLASH_BANK1 = 0x8000000, // default FLASH address
+    Flash_Partitions_FLASH_BANK2 = 0x8100000
+} Flash_Partition_Adressess;
```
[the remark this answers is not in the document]
NuclearTea: renamed as suggested, using memory region instead of partition. You are right in pointing out that generally, these areas are referred to memory regions, so it makes sense to keep it consistent

firmware/Core/Src/flash_bank/flash_bank.c
NuclearTea: its address is passed into the following function which by default, renders it mutable

firmware/Core/Src/flash_bank/flash_bank.c
NuclearTea: I left a comment explaining what page error is

firmware/Core/Src/flash_bank/flash_bank.c
NuclearTea: this is a vestige of my testing, I mainly tested with characters so this made more sense then, but in reality, we would be passing in raw integer values, so I will remove it

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
...
+        // const uint64_t double_word = *(uint64_t *)(data + (currentAddress - address));
+        uint8_t data_to_write[8] = {0};
+        memcpy(data_to_write, data + (currentAddress - address), 8);
+        const uint64_t double_word = *(uint64_t *)(data_to_write);
+        // Write the word to flash
+        status = HAL_FLASH_Program(FLASH_TYPEPROGRAM_DOUBLEWORD, currentAddress, double_word);
+        if (status != HAL_OK)
+        {
+            // If the program operation fails, break the loop
+            break;
```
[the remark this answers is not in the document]
NuclearTea: The thing is, the actions I would take to clean up are identical whether this succeeds or fails. So I don't think there is any benefit to repeating code.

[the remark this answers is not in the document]
NuclearTea: I would still have to do a:
```
if (HAL_FLASH_Lock() != HAL_OK)
    {
        return 11;
    }
```
and then return the status of the HAL_FLASH_Program(....) function

firmware/Core/Src/flash_bank/flash_bank.c
```diff
@@ -0,0 +1,124 @@
+#include "flash_bank/flash_bank.h"
+
+#include "stm32l4r5xx.h"
+#include "stm32l4xx_hal.h" // must include this before stm32l4xx_hal_flash.h
+#include "stm32l4xx_hal_flash.h"
+#include "stm32l4xx_hal_flash_ex.h"
+#include <string.h>
+/**
```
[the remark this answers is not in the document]
NuclearTea: Good point, done!

firmware/Core/Src/flash_bank/flash_bank.c
DeflateAwning: The struct is being modified? Are you sure?

firmware/STM32L4R5ZITx_FLASH.ld
NuclearTea: Not sure if you can do it using the IOC file, but I will add it in the docs in a bit

firmware/Core/Src/telecommands/flash_bank_telecommands_defs.c
NuclearTea: done

DeflateAwning:
Not all naming convention deficiencies (namespaces, etc.) were implemented properly. I will review once all requested changes are implemented.

NuclearTea:
I have fixed the returning of the stm32_interal_flash_erase driver function and I think have addressed all naming convention violations. Please let me know if I missed anything.

Thanks!

firmware/Core/Inc/stm32_internal_flash_drivers/stm32_internal_flash_drivers.h
```diff
@@ -0,0 +1,23 @@
...
+typedef enum
+{
+    STM32_INTERNAL_FLASH_MEMORY_REGION_RAM_ADDRESS = 0x20000000,
+    STM32_INTERNAL_FLASH_MEMORY_REGION_RAM_2_ADDRESS = 0x10000000,
+    STM32_INTERNAL_FLASH_MEMORY_REGION_RAM_3_ADDRESS = 0x20040000,
+    STM32_INTERNAL_FLASH_MEMORY_REGION_FLASH_ADDRESS = 0x8000000, // default boot address
+    STM32_INTERNAL_FLASH_MEMORY_REGION_GOLDEN_COPY_ADDRESS = 0x8100000
+} STM32_INTERNAL_FLASH_memory_region_addresses_t;
+
+uint8_t stm32_internal_flash_write(uint32_t address, uint8_t *data, uint32_t length);
```
DeflateAwning, line 17: Make the `STM32_` all caps

DeflateAwning (changes requested):
One tiny change, thanks!

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
```diff
@@ -0,0 +1,143 @@
...
+#include "stm32l4xx_hal_flash_ex.h"
+#include <string.h>
+
+/// @brief Writes data to the flash memory in chunks of 8 bytes.
+/// @param address Address in the flash memory where the data will be written.
+/// @param data uint8_t buffer containing the data to be written.
+/// @param length Length of the data to be written.
+/// @return 0 on success, > 0 on error, 10 if HAL_FLASH_Unlock() failed, 11 if HAL_FLASH_Lock() failed
+/// @note Currently, only allowed to write to golden copy region
+/// @note if data is not 8 bytes long, it will write the data given, and clear the rest of the data
```
DeflateAwning, line 15: This note doesn't make sense. Are you saying if it's not a multiple of 8? Less than 8? More than 8? Your example shows that it clears the data "after", not "before", I think. Maybe an example written out as an array would be more helpful.

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
```diff
@@ -0,0 +1,143 @@
...
+    {
+        return 10;
+    }
+
+    // Clear all FLASH flags before starting the operation
+    __HAL_FLASH_CLEAR_FLAG(FLASH_FLAG_ALL_ERRORS);
+
+    HAL_StatusTypeDef status = HAL_OK;
+
+    for (uint32_t currentAddress = address; currentAddress < end_address; currentAddress += 8)
```
DeflateAwning, line 41: `currentAddress` is not snake case. That's fine though.

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
```diff
@@ -0,0 +1,143 @@
...
+        // currentAddress - address is the number of bytes we have written
+        // since the beginning of the function
+        memcpy(data_to_write, data + (currentAddress - address), 8);
+
+        const uint64_t double_word = *(uint64_t *)(data_to_write);
+
+        status = HAL_FLASH_Program(FLASH_TYPEPROGRAM_DOUBLEWORD, currentAddress, double_word);
+        if (status != HAL_OK)
+        {
+            break;
```
DeflateAwning, line 57: No change required if you disagree, but I finally figure out why this `break;` bothers me so much (instead of returning from there): You generally want the reported error/exeception to be the first error that happens. 

Consider the case where `HAL_FLASH_Program` fails, and then also fails to lock. The return code will be that it failed to lock, when the useful indicator is actually that it failed `HAL_FLASH_Program` in my opinion, and then subsequently failed to lock.

DeflateAwning (approved):
Gave this a better read through. Will approve because I've been nit-picking this for a bit, but I encourage you to at least read (and ideally apply) these comments

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
NuclearTea: does the following make sense:

```
/// @note Writes data in chunks of 8 bytes.
/// Ex: Suppose we wanted to write to address 0x00, and suppose that at address 0x00, the first 8 bytes looks like the following:
/// [1,2,3,4,5,6,7,8]. If we wanted to write [25,26,27,28], it would result in the following: [25,26,27,28,0,0,0,0], clearing the rest of the bytes.
```

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
NuclearTea: I think that's really fair. 

I have a couple of alternatives:

1. First
    1. That if statement returns value (lets say 3)
    2. The calling function inspects the return value, if it is 3, it will execute the following code:
```
const uint8_t write_res = stm32_internal_flash_write(...);
if (write_res == 3) {
const HAL_StatusTypeDef lock_res = HAL_Flash_Lock();
if (lock_res != HAL_OK) {
    snprintf(....);
    return 1;
    }
snprintf(....);
return 1;
}

```
    3. This is a bit messy but not sure how else to remedy this.
2. Second
    1. The function never locks internally, rather, the calling code always must call HAL_Flash_Lock() after using this function
    2. I personally don't like this because then the API layer is not segregated but then, we don't have diverting paths for how this function Locks the Flash.

NuclearTea: With both of these suggestions, the API layer will become exposed to the calling code, so I don't know how to avoid that.

1 more thing I was able to come up with is the following: 

- Essentially, we have 4 states, either:
    - Everything is successful
    - Write fails but lock is successful
    - Write is successful but lock fails
    - Write fails and lock fails.

Currently, the approach addresses the first 3, but is missing the last one. I was wondering if you had any suggestions on restructuring the code so that I don't have to expose the internal API layer to the telecommand function

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
DeflateAwning: Looks good to me!

firmware/Core/Src/stm32_internal_flash_drivers/stm32_internal_flash_drivers.c
DeflateAwning: Will leave to you to decide! We're not playing code golf though; calling the lock function in two places isn't a sin.

NuclearTea: Yes, I just realized that I over complicated it.

firmware/STM32L4R5ZITx_FLASH.ld
DeflateAwning: It is still not in the docs. Please add to the docs, if it is necessary.

DeflateAwning (changes requested):

DeflateAwning (approved):

PR #185, opened 2024-09-20
Send arbitrary bytes to UART peripheral, and read its response

vaibhavkpr:
Implements base functionality to send arbitrary configuration commands, etc. to the peripheral (the MPI/GPS/Camera).
* Tested for only USART1 (MPI - which supports DMA reception) & UART5 (EPS - Interrupt reception) which encompass the 2 configured reception modes.
* Unable to test for the expected max transmit and receive 5kb case due to current limitations. (Reception seems to work up to 2Kbs in current tests.


DeflateAwning:
Please update the PR title to describe the changes

DeflateAwning:
Updated IOC file in 434480b to support RX interrupts!

vaibhavkpr:
Can verify that the RX interrupts are working for the enabled ports! Will leave the DMA testing for UART4 Camera for when it is implemented.

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -1046,3 +1056,194 @@ uint8_t TCMDEXEC_reboot(const char *args_str, TCMD_TelecommandChannel_enum_t tcm
...
+/// @return 0: Success, 1: Error parsing args, 2: Invalid uart port requested, 3: Error transmitting data, 4: Error receiving data
+uint8_t TCMDEXEC_peripheral_send_receive_data(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+
+    // Parse UART port argument
+    char arg_uart_port_name[10] = "";
+    const uint8_t uart_port_name_parse_result = TCMD_extract_string_arg(args_str, 0, arg_uart_port_name, 10);
+
+    // Parse hex-encoded config command (string to bytes)
+    const uint16_t args_bytes_to_send_size = (strlen(args_str) - strlen(arg_uart_port_name))/2;             // Expected size of config command byte array
```
NuclearTea, line 1075: This won't work as the hex value could be sent with/without spaces changing the actual size of the buffer, also, I think we want to send 5 Kb

NuclearTea (changes requested):

firmware/Core/Src/telecommands/telecommand_definitions.c
vaibhavkpr: Thanks for pointing that out! I have updated this bug to now support 5KBs of fixed transmission buffer size and handle the with/without spaces scenario in the latest commit.

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -1046,3 +1056,194 @@ uint8_t TCMDEXEC_reboot(const char *args_str, TCMD_TelecommandChannel_enum_t tcm
     return 0;
 }
 
+/// @brief Send artibrary configuration commands to a peripheral (MPI, GPS, CAMERA)
+/// @param args_str 
+/// - Arg 0: UART port to send data to
```
DeflateAwning, line 1061: Example? Number? Name? Both/either (like the EPS channels)

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -1046,3 +1056,194 @@ uint8_t TCMDEXEC_reboot(const char *args_str, TCMD_TelecommandChannel_enum_t tcm
...
+/// - Arg 1: Data to be sent (bytes specified in base64)
```
DeflateAwning, line 1062: Base64? You parse it in hex tho....

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -1046,3 +1056,194 @@ uint8_t TCMDEXEC_reboot(const char *args_str, TCMD_TelecommandChannel_enum_t tcm
...
+                break; 
+            }
+        }
+    }
+
+    // Send back complete response           
+    snprintf(
+        &response_output_buf[strlen(response_output_buf)],
+        response_output_buf_len - strlen(response_output_buf) - 1,
+        "Data successfully transmitted to %s port.\nResponse received (%u bytes): ", arg_uart_port_name, rx_buffer_len
```
DeflateAwning, line 1235: Remove newline here.

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -1046,3 +1056,194 @@ uint8_t TCMDEXEC_reboot(const char *args_str, TCMD_TelecommandChannel_enum_t tcm
...
+    if(uart_port_name_parse_result != 0 || bytes_to_send_parse_result !=0) {
+        snprintf(
+            response_output_buf,
+            response_output_buf_len,
+            "Error parsing uart port/data to send arg: Arg 0 Err=%d, Arg 1 Err=%d", uart_port_name_parse_result, bytes_to_send_parse_result);
+        return 1;    // Error code: Error parsing args
+    }
+    
+    // Allocate space to receive incoming response
+    const size_t rx_buffer_max_size = 5120;         // 5KBs max size for rx buffer
```
DeflateAwning, line 1090: 5 KiB *

DeflateAwning (changes requested):

vaibhavkpr:
Just wanted to provide an update on this PR:
⚠️  Currently trying to debug an issue with interrupt based reception.
✅  Changes requested above have been addressed.
✅  The functions have been cleaned up further and now utilize the new logging system with better more informative messages.

vaibhavkpr:
@DeflateAwning 
I have pushed my changes and rebased.
As for the issue, I have narrowed it down to the UART_peripheral_buffer_write_idx variables in uart_handler.c which are used to populate incoming peripheral responses and not the reception itself. The index variable increments when a byte is copied over to the peripheral UART buffer. After investigating the issue it seems like the index is updated and the buffers are populated as they should but when I try to access the index value from the telecommand function itself to get the size of the response received, it keeps showing up as 0 despite being incremented in the callbacks to the number of bytes received.

NuclearTea:
@vaibhavkpr Is there any update on the bug you identified?

vaibhavkpr:
@NuclearTea I was able to find the bug, it was a race condition between the UART handler receiving the data right after transmission before it could initially reset the reception indexes. From the initial tests, the fix seems to be working well. I am just in the process of reformatting the code a bit and once I have run some more tests, this should be good to go!

DeflateAwning:
Please fix CI run

vaibhavkpr:
@DeflateAwning As I was testing this PR post the bug fixes, I found a few more things that need to be considered and would like some advice on:

- HAL_UART_Transmit is a blocking function, which may not be efficient for supporting data transmissions up to 5 KiB. A more efficient approach would be to use HAL_UART_Transmit_DMA, which is non-blocking and operates independently of the RX configuration. The RX settings can continue to use interrupts (via HAL_UART_Receive_IT) without interference. To implement this, DMA needs to be enabled for all UART ports.

- HAL_UART_Transmit appears to transmit in chunks of 32 bytes. It's unclear if this is something we can modify or a set limitation by the STM32 hardware itself. Are the peripherals are already configured to handle this scenario?

DeflateAwning:
> HAL_UART_Transmit is a blocking function, which may not be efficient for supporting data transmissions up to 5 KiB
How long does 5000 bytes at 115200 baud take? 340 seconds, I think. That's fine.

> HAL_UART_Transmit appears to transmit in chunks of 32 bytes. It's unclear if this is something we can modify or a set limitation by the STM32 hardware itself. Are the peripherals configured to handle this scenario?
I'm not sure what you mean. If you mean that the maximum length it can transmit at once is 32 bytes, then we need to call that function multiple times back-to-back, right?

vaibhavkpr:
> How long does 5000 bytes at 115200 baud take? 340 seconds, I think. That's fine.

Should be 350 milliseconds I believe, but yeah I agree it seems reasonable for this use-case.

> > HAL_UART_Transmit appears to transmit in chunks of 32 bytes. It's unclear if this is something we can modify or a set limitation by the STM32 hardware itself. Are the peripherals configured to handle this scenario?
> 
> I'm not sure what you mean. If you mean that the maximum length it can transmit at once is 32 bytes, then we need to call that function multiple times back-to-back, right?

Not exactly, HAL_UART_Transmit is able to send the entire buffer but on the reception side it appears as if the transmission is split after every 32 bytes transmitted (which initially made be believe it was using an internal buffer with a 32 byte size limit causing this). There isn't a noticeable delay so as long as the peripherals handle reception using timeouts, I think it wont be an issue.

DeflateAwning:
Testing update from last night: After testing with a receive methodology matching that of the EPS interrupt reception, none of these issues were observed. Ensure the debugger is not running/lingering.

vaibhavkpr:
> Testing update from last night: After testing with a receive methodology matching that of the EPS interrupt reception, none of these issues were observed. Ensure the debugger is not running/lingering.

After further testing, on the LORA port with >70 bytes, I am noticing the old issue again at times where the while loop seems to exit without waiting until the timeout durations. This is with the updated implementation following the EPS interrupt reception.

`Timeout used for this test (increased to make sure the while loop atleast waits 500ms`
`RX_TIMEOUT_BEFORE_FIRST_BYTE_MS 500     // Timeout duration for first rx byte in milliseconds`
`RX_TIMEOUT_BETWEEN_BYTES_MS 700         // Timeout duration for in between rx bytes in milliseconds`

`Expected Output:`
`0000000000000+0000151270_N [TELECOMMAND:NORMAL]: Transmitted 74 byte(s) to lora.`
`0000000000000+0000152264_N [TELECOMMAND:NORMAL]: Received 75 byte(s) from lora: 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 88 99 00 11 22 33 01`
`========================= Response (duration=1102ms, err=0) =========================`

`Failed Output:`
`0000000000000+0000300901_N [TELECOMMAND:NORMAL]: Transmitted 74 byte(s) to lora.`
`0000000000000+0000301132_N [TELECOMMAND:NORMAL]: Received 18 byte(s) from lora: 00 11 22 33 44 55 66 77 88 99 00 11 22 33 44 55 66 77 `
`========================= Response (duration=324ms, err=0) =========================`

Key thing to note from the output is the response duration = 324ms (despite timeout between bytes set to 700ms to be on the safe side & python script completing the entire transmission within 300ms). 

Also, an observation I didn't notice this issue while testing the GPS or the EPS port which follows the exact implementation. The significantly lower baudrate on lora's port (9600) seems to be the differentiating factor here...

firmware/Core/Src/telecommands/telecommand_definitions.c
vaibhavkpr: `Arg 0: UART port to send data to: MPI, LORA, GPS, CAMERA, EPS (case insensitive)`

Updated to support peripheral name in a case insensitive format. Sample function call now looks like:
`CTS1+uart_send_bytes_hex(MPI,00110011...)!`

vaibhavkpr:
Good News! The issue with the LORA reception failure has been resolved (A quirky race condition was found during timeout between bytes scenario)

The code has been now successfully tested with the MPI, LORA, GPS & EPS ports to handle atleast 70 bytes of data transfer and reception. Once the camera drivers are implemented, it will need to be added to this issue.

hardware_emulators/MPI_Emulator.py
```diff
@@ -1,19 +1,40 @@
...
+
+            # Check if more than 200 milliseconds have passed since the last received data
+            if len(buffer) > 0 and (time.time() - last_receive_time) > receive_timeout:
+                # Transmission is considered complete
+                # Now echo back the entire buffer at once, not in chunks
+                response = buffer + bytes([0x01])  # Append the echo byte
                 ser.write(response)
-                print(f"Responded:  {response.hex()}")
+                ser.flush()  # Ensure the data is sent immediately
+                logger.info(f"Responded:  {response.hex()}")
```
DeflateAwning, line 29: Use prefix `MPI_COMPUTER -> OBC` instead of `Responded`

hardware_emulators/MPI_Emulator.py
```diff
@@ -1,19 +1,40 @@
...
             if ser.in_waiting > 0:
+                # Read in available incoming data
                 incoming_cmd = ser.read(ser.in_waiting)
-                print(f"Received:   {incoming_cmd.hex()}")
-                
-                # Append 1 to the telecommand data received to indicate a successful reception & echo back
-                response = incoming_cmd + bytes([0x01])
+                buffer.extend(incoming_cmd)
+                last_receive_time = time.time()  # Update the last receive time
+                logger.info(f"Received: {incoming_cmd.hex()}")
```
DeflateAwning, line 20: Use prefix `OBC -> MPI_COMPUTER: `

DeflateAwning (changes requested):
Just a few small changes. Nice work!

firmware/Core/Src/telecommands/uart_telelecommand_defs.c
```diff
@@ -0,0 +1,313 @@
...
+                    LOG_source, LOG_SEVERITY_ERROR, LOG_SINK_ALL,
+                    "Unknown response return: %u", 
+                    MPI_tx_rx_status
+                );
+                return 7;   // Error code: Unhandled error
+        }
+    }
+
+    // UART 4 Selected (CAMERA)
+    else if(strcasecmp(arg_uart_port_name, "CAMERA") == 0) {
```
DeflateAwning, line 138: You need to return and error and say that this is not implemented. Currently it just fails silently.

Some electrical team members wasted a lot of time trying to troubleshoot this UART port tonight, despite not being implemented.

Why isn't this implemented?

DeflateAwning (changes requested):

firmware/Core/Src/mpi/mpi_command_handling.c
```diff
@@ -55,65 +68,55 @@ uint8_t MPI_send_telecommand_get_response(const uint8_t *bytes_to_send, const si
...
-            // Reset UART buffer write index
-            UART_mpi_rx_buffer_write_idx = 0;
+    // Stop reception from the MPI & Reset MPI transceiver mode state, Set MPI UART mode state to previous state
+    MPI_set_transceiver_state(MPI_TRANSCEIVER_MODE_INACTIVE);
+    HAL_UART_DMAStop(UART_mpi_port_handle);
+    MPI_current_uart_rx_mode = MPI_last_uart_rx_mode;
 
-            // Validate MPI response
-            const uint8_t MPI_response_result = MPI_validate_telecommand_response(bytes_to_send, MPI_rx_buffer, bytes_to_send_len);
+    // For now log the response from the MPI //TODO: Change this after testing
```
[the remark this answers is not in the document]
vaibhavkpr: I was unclear if we just wanted to log responses or write them to a file (if there was a larger response expected from one of the peripherals for example). I can get rid of this if the scope is limited here.

firmware/Core/Inc/telecommands/telecommand_definitions.h
```diff
@@ -26,5 +26,8 @@ uint8_t TCMDEXEC_available_telecommands(const char *args_str, TCMD_TelecommandCh
 
 uint8_t TCMDEXEC_reboot(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len);
+
+uint8_t TCMDEXEC_peripheral_send_receive_data(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
```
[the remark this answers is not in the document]
vaibhavkpr: Must have missed this from the earlier iteration of the function. I have removed it.

firmware/Core/Src/mpi/mpi_command_handling.c
```diff
@@ -55,65 +68,55 @@ uint8_t MPI_send_telecommand_get_response(const uint8_t *bytes_to_send, const si
...
+    // Copy the buffer to the last received byte index & clear the UART buffer
+    for (uint16_t i = 0; i < *MPI_rx_buffer_len; i++) {
+        MPI_rx_buffer[i] = UART_mpi_buffer[i];
+        //UART_mpi_buffer[i] = 0;
```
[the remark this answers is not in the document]
vaibhavkpr: Although we clear the UART buffer before recording new responses, I was thinking we could clear them after they are read from here as well as part of a safe measure here. If it would be unnecessarily redundant I can remove it.

firmware/Core/Src/telecommands/uart_telelecommand_defs.c
```diff
@@ -0,0 +1,313 @@
...
+/// @param args_str 
+/// - Arg 0: UART port to send data to: MPI, LORA, GPS, CAMERA, EPS (case insensitive)
+/// - Arg 1: Data to be sent (bytes specified as hex - Max 5KiB buffer)
+/// @param tcmd_channel The channel on which the telecommand was received, and on which the response should be sent
+/// @param response_output_buf The buffer to write the response to
+/// @param response_output_buf_len The maximum length of the response_output_buf (its size)
+/// @return 0: Success, 1: Error parsing args, 2: Invalid uart port requested, 3: Error transmitting data, 
+///         4: Error receiving data, 5: Timeout waiting for response / No response, 6: Peripheral specific error, 7: Unhandled error
+/// @note This function doesn't support toggling the EPS to send power to the peripherals
+/// @todo Add support to send data to the camera once it's supporting functions are implemented
```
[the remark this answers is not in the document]
vaibhavkpr: Its inside as well, I'll get rid of the redundancy here.

firmware/Core/Src/telecommands/uart_telelecommand_defs.c
vaibhavkpr: For the camera since it utilizes DMA very much like the MPI, I had planned to implement it the same way as the MPI with a modularized send/receive function once the Camera code was added to main. My main reasoning for this was to avoid multiple implementations of the same send/receive functionality.

My bad on missing the error message earlier—it’s definitely crucial here.

I can add the basic TX functionality here if that is preferred, I was just a bit optimistic seeing the progress on the Camera code that I decided to just wait on its push before adding it.

DeflateAwning:
When I used this with the EPS the other day, it did not work. It received a few bytes, and then stopped after that. We must check on this.

vaibhavkpr:
> When I used this with the EPS the other day, it did not work. It received a few bytes, and then stopped after that. We must check on this.

That's odd, I'll have a look and see what I can find out about this.

vaibhavkpr:
> A little while back, we found a bug and localized it to the maximum length of the LOG_message payload.
> 
> It would be great to get this merged soon. Let me know when this is fixed by doing multiple logging calls with chunks of the UART payload.

Updated code has been pushed! Response is now logged in chunks of 256 characters. 
Note: If the max size defined in log.c as "LOG_FORMATTED_MESSAGE_MAX_LENGTH" can be added to the header file, it can be referenced instead of being hardcoded.

DeflateAwning:
Feel free to move `LOG_FORMATTED_MESSAGE_MAX_LENGTH` as is beneficial!

vaibhavkpr:
> Feel free to move `LOG_FORMATTED_MESSAGE_MAX_LENGTH` as is beneficial!

Change has been made. Should be good to go now. 
Awaiting #247 before being merged.

DeflateAwning:
Now that the GPS code is merged, can you please look at this and work to rebase it, leaving the GPS UART handler content as it is currently? Thanks!

DeflateAwning:
Will be tested and merged in https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/pull/272.

PR #192, opened 2024-09-21
FreeRTOS Thread to monitor high stack watermarks

### LINKED ISSUE #128: Create FreeRTOS thread which monitors the high stack watermarks and logs warnings if any tasks exceed thresholds

christopherkinyua:
## Related Issues

Closes #128 

## Summary

This PR creates an osThread that will periodically check the high stack watermark to see if it is below a defined threshold. This is done to note when a potential stack overflow may occur.

## Changes Made

The changes made to realize this were as follows:
- Created a thread called `TASK_monitor_freertos_highstack_watermarks` in `rtos_tasks.c` and added the function definition in `rtos_tasks.h`
- Created a osThreadId_t,  osThreadAttr_t and thread for `TASK_monitor_freertos_highstack_watermarks`  in main

## Special Notes for Your Reviewer(s)
During the implementation of the above steps, I noticed that each task has varying stack sizes ranging from 256-8192. Therefore my `highstack_watermark_threshold` may not be large enough for specific tasks like `TASK_execute_telecommands_Handle`. I was aiming to have a threshold value of 20% for their allocated stack size.

I was thinking of creating a wrapper function, let's say called `createThread`, that will be used to create the threads so I can dynamically allocate a threshold value. I'd create a struct called `TaskInfo_t` that contains the task name and the stack size. I would then have a global array of the struct which would be populated during the creation of the thread. Do a string comparison of each task name and then dynamically allocate a 20% threshold for each task based on their stack size.

I was wondering if this was a little too complex for the simplicity of this ticket. Let me know what you think, otherwise, I can just modify the threshold value that I already have in place to be something higher than 50, which is what I have right now.

### DIFF
--- a/firmware/Core/Src/main.c
+++ b/firmware/Core/Src/main.c
@@ -109,9 +109,34 @@ const osThreadAttr_t TASK_execute_telecommands_Attributes = {
 osThreadId_t TASK_monitor_freertos_highstack_watermarks_Handle;
 const osThreadAttr_t TASK_monitor_freertos_highstack_watermarks_Attributes = {
   .name = "TASK_monitor_freertos_highstack_watermarks",
-  .stack_size = 512,
-  .priority = (osPriority_t) osPriorityNormal,
+  .stack_size = 1024,
+  .priority = (osPriority_t) osPriorityBelowNormal6,
 };
+
+const Task_Info_t task_handles_array [] = {
+  {
+    .task_handle = &defaultTaskHandle,
+    .stack_size_bytes = 128 * 4
+  },
+  {
+    .task_handle = &TASK_DEBUG_print_heartbeat_Handle,
+    .stack_size_bytes = 256
+  },
+  {
+    .task_handle = &TASK_handle_uart_telecommands_Handle,
+    .stack_size_bytes = 8192
+  },
+  {
+    .task_handle = &TASK_execute_telecommands_Handle,
+    .stack_size_bytes = 8192
+  },
+  {
+    .task_handle = &TASK_monitor_freertos_highstack_watermarks_Handle,
+    .stack_size_bytes = 1024
+  },
+};
+
+const uint32_t task_handles_array_size = sizeof(task_handles_array)/ sizeof(Task_Info_t);
 
 
 /* USER CODE END PV */
--- a/firmware/Core/Src/main.c
+++ b/firmware/Core/Src/main.c
@@ -116,23 +116,23 @@ const osThreadAttr_t TASK_monitor_freertos_highstack_watermarks_Attributes = {
 const Task_Info_t task_handles_array [] = {
   {
     .task_handle = &defaultTaskHandle,
-    .stack_size_bytes = 128 * 4
+    .task_attribute = &defaultTask_attributes
   },
   {
     .task_handle = &TASK_DEBUG_print_heartbeat_Handle,
-    .stack_size_bytes = 256
+    .task_attribute = &TASK_DEBUG_print_heartbeat_Attributes
   },
   {
     .task_handle = &TASK_handle_uart_telecommands_Handle,
-    .stack_size_bytes = 8192
+    .task_attribute = &TASK_handle_uart_telecommands_Attributes
   },
   {
     .task_handle = &TASK_execute_telecommands_Handle,
-    .stack_size_bytes = 8192
+    .task_attribute = &TASK_execute_telecommands_Attributes
   },
   {
     .task_handle = &TASK_monitor_freertos_highstack_watermarks_Handle,
-    .stack_size_bytes = 1024
+    .task_attribute = &TASK_monitor_freertos_highstack_watermarks_Attributes
   },
 };
 
--- a/firmware/Core/Src/main.c
+++ b/firmware/Core/Src/main.c
@@ -106,6 +106,13 @@ const osThreadAttr_t TASK_execute_telecommands_Attributes = {
   .priority = (osPriority_t) osPriorityNormal,
 };
 
+osThreadId_t TASK_service_eps_watchdog_Handle;
+const osThreadAttr_t TASK_service_eps_watchdog_Attributes = {
+  .name = "TASK_service_eps_watchdog",
+  .stack_size = 512, //in bytes
+  .priority = (osPriority_t) osPriorityNormal, //TODO: Figure out which priority makes sense for this task
+};
+
 osThreadId_t TASK_monitor_freertos_highstack_watermarks_Handle;
 const osThreadAttr_t TASK_monitor_freertos_highstack_watermarks_Attributes = {
   .name = "TASK_monitor_freertos_highstack_watermarks",
--- a/firmware/Core/Src/main.c
+++ b/firmware/Core/Src/main.c
@@ -120,26 +120,36 @@ const osThreadAttr_t TASK_monitor_freertos_highstack_watermarks_Attributes = {
   .priority = (osPriority_t) osPriorityBelowNormal6,
 };
 
-const Task_Info_t task_handles_array [] = {
+Task_Info_t task_handles_array [] = {
   {
     .task_handle = &defaultTaskHandle,
-    .task_attribute = &defaultTask_attributes
+    .task_attribute = &defaultTask_attributes,
+    .lowest_highstack_watermark_bytes = UINT32_MAX
   },
   {
     .task_handle = &TASK_DEBUG_print_heartbeat_Handle,
-    .task_attribute = &TASK_DEBUG_print_heartbeat_Attributes
+    .task_attribute = &TASK_DEBUG_print_heartbeat_Attributes,
+    .lowest_highstack_watermark_bytes = UINT32_MAX
   },
   {
     .task_handle = &TASK_handle_uart_telecommands_Handle,
-    .task_attribute = &TASK_handle_uart_telecommands_Attributes
+    .task_attribute = &TASK_handle_uart_telecommands_Attributes,
+    .lowest_highstack_watermark_bytes = UINT32_MAX
   },
   {
     .task_handle = &TASK_execute_telecommands_Handle,
-    .task_attribute = &TASK_execute_telecommands_Attributes
+    .task_attribute = &TASK_execute_telecommands_Attributes,
+    .lowest_highstack_watermark_bytes = UINT32_MAX
+  },
+  {
+    .task_handle = &TASK_service_eps_watchdog_Handle,
+    .task_attribute = &TASK_service_eps_watchdog_Attributes,
+    .lowest_highstack_watermark_bytes = UINT32_MAX
   },
   {
     .task_handle = &TASK_monitor_freertos_highstack_watermarks_Handle,
-    .task_attribute = &TASK_monitor_freertos_highstack_watermarks_Attributes
+    .task_attribute = &TASK_monitor_freertos_highstack_watermarks_Attributes,
+    .lowest_highstack_watermark_bytes = UINT32_MAX
   },
 };
 
--- a/firmware/Core/Src/main.c
+++ b/firmware/Core/Src/main.c
@@ -109,7 +109,7 @@ const osThreadAttr_t TASK_execute_telecommands_Attributes = {
 osThreadId_t TASK_monitor_freertos_highstack_watermarks_Handle;
 const osThreadAttr_t TASK_monitor_freertos_highstack_watermarks_Attributes = {
   .name = "TASK_monitor_freertos_highstack_watermarks",
-  .stack_size = 256,
+  .stack_size = 512,
   .priority = (osPriority_t) osPriorityNormal,
 };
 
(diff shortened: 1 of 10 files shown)


firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -163,3 +163,39 @@ void TASK_execute_telecommands(void *argument) {
...
+			continue;
+		}
+
+		//Check the highstack watermarks for each task
+		for(UBaseType_t x = 0; x < number_of_tasks; x++){
+			if(task_statuses[x].usStackHighWaterMark < highstack_watermark_threshold)
+			{
+				LOG_message(
+					LOG_SYSTEM_OBC, LOG_SEVERITY_WARNING, LOG_SINK_ALL,
+					"Warning: Task %s is nearing stack overflow. Remaining stack size is: %u bytes.",
```
DeflateAwning, line 192: Semantic: "is nearing" suggests that, at the exact moment the message is being printed, the situation is that Task X has lots of memory usage. My understanding is that the high watermark is a "logged" metric, and represents the "worst" way it's ever been (and not the way it "currently is").

Perhaps change to `"Task '%s' approached a stack overflow. Worst remaining stack size was: %u bytes."`, or similar.

DeflateAwning (changes requested):
This looks super awesome! Probably could merge it as-is, but if you're interested in a few changes:

Just a few tweaks and opportunities to go a bit above what the Issue originally mentioned.

Inline with what you were mentioning in the PR description, I think a percentage-based threshold would be awesome!

I wonder if it would work to populate a global extern array of pointers to the task handles in `main.c`; then, I think you'd be able to iterate through that array, and call [`uxTaskGetStackHighWaterMark`](https://www.freertos.org/Documentation/02-Kernel/04-API-references/03-Task-utilities/04-uxTaskGetStackHighWaterMark) directly without creating the array here like you currently do. I think that you'd be able to look up the attributes to figure out the full max stack size too.

I may be totally wrong about some of that; I think it's a simpler direction to check out first before we create a function that registers the task handles (if I understood your suggestion right).

firmware/Core/Src/rtos_tasks/rtos_tasks.c
christopherkinyua: Done

firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -163,3 +163,39 @@ void TASK_execute_telecommands(void *argument) {
 
 	} /* End Task's Main Loop */
 }
+
+void TASK_monitor_freertos_highstack_watermarks(void *argument) {
+	TASK_HELP_start_of_task();
+
+	const uint32_t highstack_watermark_threshold = 50;
+	uint32_t total_run_time;
```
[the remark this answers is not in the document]
christopherkinyua: Done

firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -163,3 +163,39 @@ void TASK_execute_telecommands(void *argument) {
 
 	} /* End Task's Main Loop */
 }
+
+void TASK_monitor_freertos_highstack_watermarks(void *argument) {
+	TASK_HELP_start_of_task();
+
+	const uint32_t highstack_watermark_threshold = 50;
```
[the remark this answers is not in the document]
christopherkinyua: Done

christopherkinyua:
I've implemented a percentage-based dynamic threshold value for the task handles. I was having difficulties determining the task stack size from the task handles so I made a struct to contain a pointer to the task handle and the stack size.

I am also dereferencing the task handles in the thread so I was wondering if I could store the task handle itself instead of its pointer in the struct `Task_info_t` as the functions `uxTaskGetStackHighWaterMark` and `pcTaskGetName` do not take pointer values.

firmware/Core/Src/main.c
```diff
@@ -106,6 +106,37 @@ const osThreadAttr_t TASK_execute_telecommands_Attributes = {
...
+const osThreadAttr_t TASK_monitor_freertos_highstack_watermarks_Attributes = {
+  .name = "TASK_monitor_freertos_highstack_watermarks",
+  .stack_size = 1024,
+  .priority = (osPriority_t) osPriorityBelowNormal6,
+};
+
+const Task_Info_t task_handles_array [] = {
+  {
+    .task_handle = &defaultTaskHandle,
+    .stack_size_bytes = 128 * 4
```
DeflateAwning, line 119: Instead of duplicating the stack size values here, what do you think about storing a pointer to the attributes struct and reading it from there? I think that might be a bit cleaner?

firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -166,3 +167,52 @@ void TASK_execute_telecommands(void *argument) {
...
+				}	
+			}
+		}
+		osDelay(5000);
+
+	} /* End Task's Main Loop */
+}
+
+
+		// uint32_t total_run_time;
```
DeflateAwning, line 205: Avoid commented-out code

firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -166,3 +167,52 @@ void TASK_execute_telecommands(void *argument) {
...
+
+			if(task_handles_array[x].task_handle != NULL)
+			{
+				// Dereferencing the task_handle pointer
+				osThreadId_t dereferenced_task_handle = *(task_handles_array[x].task_handle);
+
+				// Determine the threshold of the task
+				uint32_t task_threshold_words = (task_handles_array[x].stack_size_bytes * CONFIG_highstack_watermark_percentage_threshold) / 400;
+
+				// Get the highstack watermark
```
DeflateAwning, line 186: Is this going to constanty spew warnings after it passes the threshold? 

Might be fine, but might be nice if there was any easy way to only warn if it gets even worse (or something like that).

firmware/Core/Src/rtos_tasks/rtos_tasks.c
```diff
@@ -166,3 +167,52 @@ void TASK_execute_telecommands(void *argument) {
...
+
+		for(uint32_t x = 0; x < task_handles_array_size; x++){
+
+			if(task_handles_array[x].task_handle != NULL)
+			{
+				// Dereferencing the task_handle pointer
+				osThreadId_t dereferenced_task_handle = *(task_handles_array[x].task_handle);
+
+				// Determine the threshold of the task
+				uint32_t task_threshold_words = (task_handles_array[x].stack_size_bytes * CONFIG_highstack_watermark_percentage_threshold) / 400;
```
DeflateAwning, line 184: What is magic number 400?

I'd recommend converting to bytes and working with values and percentages in bytes instead of words.

DeflateAwning (changes requested):
We should check that `usStackHighWaterMark` (in the telecommand) is measured in bytes and not words.

firmware/Core/Src/rtos_tasks/rtos_tasks.c
christopherkinyua: Totally forgot that was there. Thanks

christopherkinyua:
> Unrelated thought: We should check that `usStackHighWaterMark` (in the telecommand) is measured in bytes and not words.

In the FreeRTOS docs, the return value of  `usStackHighWaterMark` is the high water mark in words. Therefore, Ill have to make some changes to the telecommands that use this. I can address it in this ticket if you'd like.

firmware/Core/Src/main.c
christopherkinyua: Sure thing, let me attempt this approach

firmware/Core/Src/rtos_tasks/rtos_tasks.c
christopherkinyua: I was doing the comparison in words as that was the return value of `uxTaskGetStackHighWaterMark()` hence dividing by a factor of 4. I divided by 100 as the `CONFIG_highstack_watermark_percentage_threshold` is 20 instead of 0.2

Ill move forward in performing the comparison in bytes

firmware/Core/Src/rtos_tasks/rtos_tasks.c
christopherkinyua: Thats true.

So we need to keep track of when and the remaining stack based on the `task_highstack_watermark_bytes` and do a comparison between `task_highstack_watermark_bytes` and let us say `last_reported_task_highstack_watermark_bytes`

If the `task_highstack_watermark_bytes` is smaller than `last_reported_task_highstack_watermark_bytes` then we print to UART the warning?

christopherkinyua: image

For the `TASK_DEBUG_print_heartbeat` its worst remaining stack size is always 0 so, it will log a warning only once if I set the condition to be less than `last_reported_task_highstack_watermark_bytes` and always if I set it to be less than or equal to `last_reported_task_highstack_watermark_bytes`

christopherkinyua: If you have an idea of how to go about it, I am open to suggestions. Otherwise, for now, I could move forward with the idea I put above.

christopherkinyua: Done

christopherkinyua:
@DeflateAwning I addressed the PR comments. I decided to modify the `Task_Info_t` struct to contain `lowest_highstack_watermark_bytes`. This would help me keep track of the task handle and its lowest stack size was the challenge for me. 

Anyway, the PR is ready for your review.

DeflateAwning:
Tacked on some changes in a new commit! 

The following testing demonstrates this feature successfully!

```
CTS1+freertos_demo_stack_usage(5000)!
========================= UART Telecommand Received =========================
CTS1+freertos_demo_stack_usage(5000)!
=========================
Parsed telecommand (len=37): 'CTS1+freertos_demo_stack_usage(5000)!'
0000000000000+0000059287_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'freertos_demo_stack_usage'=========================
========================= Response (duration=14ms, err=0) =========================

===========================================================================
FrontierSat time: 19700101T00:00:59.854, Uptime: 59854 ms
FrontierSat time: 19700101T00:01:00.849, Uptime: 60849 ms
FrontierSat time: 19700101T00:01:01.844, Uptime: 61844 ms
0000000000000+0000062216_N [OBC:WARNING]: Warning: Task 'TASK_execute_telecommands' approached a stack overflow. Worst remaining stack size was: 880 bytes.
FrontierSat time: 19700101T00:01:02.839, Uptime: 62839 ms
FrontierSat time: 19700101T00:01:03.834, Uptime: 63834 ms
FrontierSat time: 19700101T00:01:04.829, Uptime: 64829 ms
CTS1+freertos_demo_stack_usage(6000)!
========================= UART Telecommand Received =========================
CTS1+freertos_demo_stack_usage(6000)!
=========================
Parsed telecommand (len=37): 'CTS1+freertos_demo_stack_usage(6000)!'
0000000000000+0000065708_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'freertos_demo_stack_usage'=========================
========================= Response (duration=16ms, err=0) =========================

===========================================================================
FrontierSat time: 19700101T00:01:05.824, Uptime: 65824 ms
FrontierSat time: 19700101T00:01:06.819, Uptime: 66819 ms
0000000000000+0000067234_N [OBC:WARNING]: Warning: Task 'TASK_execute_telecommands' approached a stack overflow. Worst remaining stack size was: 0 bytes.
FrontierSat time: 19700101T00:01:07.814, Uptime: 67814 ms
FrontierSat time: 19700101T00:01:08.809, Uptime: 68809 ms
FrontierSat time: 19700101T00:01:09.804, Uptime: 69804 ms
CTS1+freertos_demo_stack_usage(7000)!
OBC->EPS DATA (no tags): 1A 07 06 00 
OBC->EPS DATA (with tags): 3C 63 6D 64 3E 1A 07 06 00 3C 2F 63 6D 64 3E 
EPS->OBC ERROR: timeout before first byte received
0000000000000+0000070418_N [EPS:ERROR]: EPS_CMD_watchdog() -> Error: 3
========================= UART Telecommand Received =========================
CTS1+freertos_demo_stack_usage(7000)!
=========================
Parsed telecommand (len=37): 'CTS1+freertos_demo_stack_usage(7000)!'
0000000000000+0000070529_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'freertos_demo_stack_usage'=========================
========================= Response (duration=19ms, err=0) =========================

===========================================================================
FrontierSat time: 19700101T00:01:10.799, Uptime: 70799 ms
FrontierSat time: 19700101T00:01:11.794, Uptime: 71794 ms
```

DeflateAwning (dismissed):
Please give this a look over with my new small changes! Feel free to let me know if there's anything you think we should do differently, or if you find that it works notably differently now.

Feel free to merge once you're happy with it!

DeflateAwning (approved):

christopherkinyua:
Gave this a look and I like the implementation of the issue. Works as I intended it to. I didn't consider to perform a safety check at the beginning. I'll make sure to add that in future tickets.

christopherkinyua:
@DeflateAwning I just noticed that after I called the `CTS1+freertos_demo_stack_usage(7000)!` command, the program halted and required a flash to get to work.

I think it is an unintended bug and I wanted to see if you were getting the same.

To replicate the scenario:
- Flash the stm with the code
- Call telecommand `CTS1+freertos_demo_stack_usage(7000)!`
- Wait for 10 seconds

DeflateAwning:
The `TASK_execute_telecommands` thread has `stack_size = 8192`. It doesn't seem unreasonable that trying to "allocate" a 7000 byte variable-length array (VLA) on the stack would cause a stack overflow and crash that thread (which crashes the whole RTOS). I believe I saw the same behaviour. 

It's interesting that you note that it happened after a delay though.

Feel free to create an issue for it and we can dig into it further! If you do, make sure to include the reproduction example. Would be interesting to monitor each thread's stats leading up to the crash.

PR #193, opened 2024-09-22
Boom experiment drivers and telecommands

### LINKED ISSUE #96: Experiment Boom: create deployment commands, configuration, etc.
Create system to deploy experiment boom.

Should turn power on to the deployment mechanism, monitor power usage, and automatically turn the boom deploy GPIO off after a timeout (or when commanded), or on lower battery power (battery power monitor may be monitored elsewhere).

DeflateAwning:
Create drivers and telecommand(s) for the Deployable Composite Lattice Boom.

Fixes #96.

### DIFF
--- a/firmware/Core/Inc/boom_deploy_drivers/boom_deploy_drivers.h
+++ b/firmware/Core/Inc/boom_deploy_drivers/boom_deploy_drivers.h
@@ -1,0 +1,13 @@
+#ifndef INCLUDE_GUARD__BOOM_DEPLOY_DRIVERS_H
+#define INCLUDE_GUARD__BOOM_DEPLOY_DRIVERS_H
+
+
+#include "main.h"
+#include <stdint.h>
+
+
+uint8_t BOOM_get_pgood_status();
+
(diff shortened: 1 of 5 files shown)


DeflateAwning:
Note: Still requires testing. Will update when testing is completed.

firmware/Core/Inc/boom_deploy_drivers/boom_deploy_drivers.h
```diff
@@ -0,0 +1,13 @@
+#ifndef INCLUDE_GUARD__BOOM_DEPLOY_DRIVERS_H
+#define INCLUDE_GUARD__BOOM_DEPLOY_DRIVERS_H
+
+
+#include "main.h"
```
NuclearTea, line 5: Why do you need to include main.h here? You already include it in the .c file, not sure why you need it here?

DeflateAwning: Ah good catch, thanks

PR #206, opened 2024-10-04
Automated Python Unit Testing Framework

kev9268:
Added unit tests for hello world, echo args and for delay ms.
Includes a function that reads and compares expected response line and input line


DeflateAwning:
This looks really neat so far! Very well done! :) 

I quite like how simple and clean the fixture part is! I'm excited to keep expanding on this! Will give it a bit more review later on.

NuclearTea:
@DeflateAwning Need some clarification on what else to add in order to complete #5 , this PR.

firmware/.env
NuclearTea, line 1: Might be a good idea to prefix all of these variables with something like: "PY_TEST_...".

It could be possible that we need to use other environment variables that represent similar things, this way, you have the name distinguish them

NuclearTea (changes requested):

NuclearTea (approved):
Nice!

PR #216, opened 2024-10-20
LFS Telecommand to list files / directories [Issue #7]

Saksham-P:
Added Telecommand to create directory
- Arg1: Directory Name

Added Telecommand to list directory entries
- Arg1: Root Directory
- Arg2: Offset of how many files to skip at the beginning (including `.` and `..`)
- Arg3: Number of files to display in the listing


Saksham-P:
The implementation for the listing directory telecommand works as intended, although I just want to make sure if the implementation is as optimized and efficient as it can be.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -110,56 +108,80 @@ int8_t LFS_unmount()
...
+    if (count == 0) {
+        count = -1;
+    }
+
+    // result is positive on success, 0 at the end of directory, or negative on failure.
     int8_t read_dir_result = 1;
-    while (read_dir_result >= 0)
+    struct lfs_info info;
+    DEBUG_uart_print_str("Name \t bytes\n");
+    while (read_dir_result > 0)
```
NuclearTea, line 161: Regarding this while loop, there is a small thing that is occurring with the current code flow. if lfs_dir_read returns an error, you have to go through the entire loop before breaking and it could actually be weird behaviour when you try to print the information. 

Alternatives:
since count and read_dir_result both need to be valid, you could do something like:

```
for (read_dir_result = lfs_dir_read(...); read_dir_result > 0 && count != 0; count--) {
// check offset

// check info.type

}
```
You could also do a while loop which is just true, and breaks if lfs_dir_read is invalid

Lastly, you could also try a do-while loop if you really want.

Just some ideas, go with what you think makes the most sense.

NuclearTea (commented):
Nice work Overall, just some considerations.

NuclearTea:
@Saksham-P @KaleF07 Wondering if the comments made sense?

KaleF07:
> @Saksham-P @KaleF07 Wondering if the comments made sense?

Saksham was mostly the one working on this issue, some of my commits are from previous LFS issues. I will let him respond to the comments as I have not looked at the code yet

Saksham-P:
I updated the code, and didn't change the loop, but just checked if we read dir gave an error, we break instead of returning the error, and then return the error after closing the directory.

firmware/Core/Src/littlefs/littlefs_helper.c
Saksham-P: I didn't make any major changes to the loop structure, rather just moved the if statement

firmware/Core/Src/telecommands/lfs_telecommand_defs.c
```diff
@@ -48,6 +47,84 @@ uint8_t TCMDEXEC_fs_unmount(const char *args_str, TCMD_TelecommandChannel_enum_t
...
+        return 1;
+    }
+
+    int8_t list_directory_result = LFS_list_directory(arg_root_directory_path, (uint16_t) arg_listing_offset, (int16_t) arg_listing_count);
+    if (list_directory_result < 0) {
+        snprintf(response_output_buf, response_output_buf_len, "LittleFS List Directory Error: %d\n", list_directory_result);
+        return 1;
+    }
+    
+    // snprintf(response_output_buf, response_output_buf_len, "LittleFS Successfully Unounted!\n");
```
NuclearTea, line 97: If this is not needed, we can remove this commented out code

NuclearTea (commented):
Looks great! Just need to confirm whether adding the delete file TCMD to this PR makes sense. Otherwise, great job!

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -110,124 +108,162 @@ int8_t LFS_unmount()
...
 
 /**
- * @brief Creates directory
- * @param dir_name Pointer to cstring holding the name of the directory
+ * @brief Removes / deletes the file specified
+ * @param file_name Pointer to cstring holding the file name to remove
  * @retval 0 on success, 1 if LFS is unmounted, negative LFS error codes on failure
  */
-int8_t LFS_make_directory(const char dir_name[])
+int8_t LFS_delete_file(const char file_name[])
```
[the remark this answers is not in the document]
parker-research: Either option seems good! If the code is here, probably makes sense to do it here.

Saksham-P:
I've replaced as many debug statements to LOG_message statements, there are still a few that I intentionally left in the code as they are used to send usefull information to the Serial, for example when listing the directories to the serial, it uses Debug_UART.

I've also left the `LFS_delete_file` out of this commit, as I saw Kale is already working on that issue for now, I will catch up with him to make sure we are on the same page once going through that.

Makiv1:
As per @NuclearTea 's request, I have tested this branch and the code seems to be working correctly.
First I ran `fs_list_directory(,0,10)` which returned the following:
```
========================= UART Telecommand Received =========================
CTS1+fs_list_directory(,0,10)!

=========================
Parsed telecommand (len=32): 'CTS1+fs_list_directory(,0,10)!
'
0000000000000+0000233437_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'fs_list_directory'=========================
Name 	 bytes
.
..
lifecycle

Successfully Listed Directory Contents.
0000000000000+0000233608_N [LFS:NORMAL]: Successfully listed contents from directory: 
========================= Response (duration=158ms, err=0) =========================

===========================================================================

```
This lines up with my expectation as I have created a directory called lifecycle while working on another issue.

Then running `fs_make_directory(test_dir)` **twice** gives the following:
```
Parsed telecommand (len=35): 'CTS1+fs_make_directory(test_dir)!
'
0000000000000+0000524074_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'fs_make_directory'=========================
0000000000000+0000524580_N [LFS:NORMAL]: Successfully created directory: test_dir
========================= Response (duration=494ms, err=0) =========================

===========================================================================
FrontierSat time: 19700101T00:08:44.603, Uptime: 524603 ms
FrontierSat time: 19700101T00:08:45.598, Uptime: 525598 ms
FrontierSat time: 19700101T00:08:46.593, Uptime: 526593 ms
FrontierSat time: 19700101T00:08:47.588, Uptime: 527588 ms
========================= UART Telecommand Received =========================
CTS1+fs_make_directory(test_dir)!

=========================
Parsed telecommand (len=35): 'CTS1+fs_make_directory(test_dir)!
'
0000000000000+0000528223_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'fs_make_directory'=========================
0000000000000+0000528353_N [LFS:WARNING]: Directory test_dir already exists.
========================= Response (duration=117ms, err=1 !!!!!! ERROR !!!!!!) =========================
LittleFS Make Directory Error: -17

===========================================================================
```  

The first time, directory creation succeeds and is confirmed by calling `fs_list_directory(,0,10)`. The second time it fails as the directory already exists. This seems correct.

Finally by calling `fs_list_directory(lifecycle,0,10)` we list the folders inside of this directory and get:
```
========================= UART Telecommand Received =========================
CTS1+fs_list_directory(lifecycle,0,10)!

=========================
Parsed telecommand (len=41): 'CTS1+fs_list_directory(lifecycle,0,10)!
'
0000000000000+0000875407_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0.
========================= Executing telecommand 'fs_list_directory'=========================
Name 	 bytes
.
..
deploy_antenna_on_boot_enabled.bool	1 bytes

Successfully Listed Directory Contents.
0000000000000+0000875775_N [LFS:NORMAL]: Successfully listed contents from directory: lifecycle
========================= Response (duration=357ms, err=0) =========================

===========================================================================
```

This also seems correct as I have created deploy_antenna_on_boot_enabled.bool while working on another issue.

I have also tried listing a directory which didn't exist: This Failed with error -2.

### Some Closing Thoughts
* code seems to be working correctly when tested on some common scenarios
* This is super useful for when working with the file system!
*  perhaps the response could be a bit more descriptive (logs could label which entries are files and which are directories. errors could be a bit more verbose, avoiding the need for error code lookups in the lfs.h file). Although the code in its current state resolves the linked issue

Great work this is super sweet! Hope this helps!

NuclearTea (commented):
LGTM! @DeflateAwning wondering what you think. The code has been tested by @Makiv1 and has passed testing.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -108,126 +108,156 @@ int8_t LFS_mount()
...
+        }
 
-    DEBUG_uart_print_str("Successfully Listed Directory Contents.\n");
+        if (count == 0) {
+            break;
+        } else {
+            count--;
+        }
+
+        DEBUG_uart_print_str(info.name);
```
DeflateAwning, line 183: Please switch all this to `LOG_message`

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -108,126 +108,156 @@ int8_t LFS_mount()
...
+            DEBUG_uart_print_uint32(info.size);
+            DEBUG_uart_print_str(" bytes");
+        }
+        DEBUG_uart_print_str("\n");
+    }
 
     int8_t close_dir_result = lfs_dir_close(&LFS_filesystem, &dir);
     if (close_dir_result < 0)
     {
         DEBUG_uart_print_str("Error closing directory.\n");
```
DeflateAwning, line 196: Please remove this DEBUG_uart_print when you add LOG_message calls

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -108,126 +108,156 @@ int8_t LFS_mount()
...
+        LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_CRITICAL, LOG_all_sinks_except(LOG_SINK_FILE), "Error closing directory: %s", root_directory);
         return close_dir_result;
     }
 
-    return 0;
+    if (read_dir_result < 0) {
+        return read_dir_result;
+    } else {
+        DEBUG_uart_print_str("Successfully Listed Directory Contents.\n");
```
DeflateAwning, line 204: Please remove DEBUG_uart_print_str. Please try to find all instances in the new code you're adding here.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -57,22 +55,22 @@ struct lfs_file_config LFS_file_cfg = {
...
 {
-	int8_t result = lfs_format(&LFS_filesystem, &LFS_cfg);
-	if (result < 0)
-	{
-		DEBUG_uart_print_str("Error formatting!\n");
-		return result;
+	int8_t format_result = lfs_format(&LFS_filesystem, &LFS_cfg);
+	if (format_result < 0) {
+        LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_CRITICAL, LOG_all_sinks_except(LOG_SINK_FILE), "Error formatting FLASH memory!");
+		return format_result;
```
DeflateAwning, line 68: Can you please convert this file's indentation to use spaces instead of tabs. There's a few places it's mixed in this repo, but in general, we prefer spaces. Ideally, never mix in the same file.

DeflateAwning (changes requested):
This looks fantastic, thanks! Just a few tiny changes, but awesome overall.

In addition to the review comments, in the display, can we please suffix folders (but not files) with `/`. So, for example, it would look like this:
```
========================= Executing telecommand 'fs_list_directory'=========================
Name 	 bytes
./
../
lifecycle/
some_file.txt  12
```

Saksham-P:
### Update:
image
Above image displays how the directory is listed. Got rid of all Debugs in `littlefs_helper.c`. Removed `static_config.h` include as well as `debug_uart.h` include. Could implement sorting in the future based on a criteria.

### Concern?
After pushing my changes, I realized that my local build doesn't give an error or warning when I am using `%d` to print `uint32_t`, but the build on GitHub does. (littlefs_helper.c, Line 178)
- `LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_NORMAL, LOG_all_sinks_except(LOG_SINK_FILE), "%s\t%ld bytes", info.name, info.size);`

My current arm compiler is `arm-none-eabi-gcc\13.2.1-1.1.1` which may be the issue?

DeflateAwning:
After merging/working on the version in `main`, it should warn when you use an invalid string on `LOG_message` now. Was a recent improvement in #238

NuclearTea:
@Saksham-P Have you had a chance to take a look at the comment parker left

Saksham-P:
> @Saksham-P Have you had a chance to take a look at the comment parker left

Yea, I had updated my branch from main and the warning do show in the build locally. Regardless, the current version of the code doesn't have any type warnings, so it is good to go!

firmware/Core/Src/telecommands/lfs_telecommand_defs.c
```diff
@@ -48,6 +47,82 @@ uint8_t TCMDEXEC_fs_unmount(const char *args_str, TCMD_TelecommandChannel_enum_t
...
 
+/// @brief Telecommand: List all the files and directories within a given directory
+/// @param args_str
+/// - Arg 0: Root Directory path as string
+/// - Arg 1: (Offset) Number of entries to skip at the beginning
+/// - Arg 2: (Count) Number entries to display
+uint8_t TCMDEXEC_fs_list_directory(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+
+    char arg_root_directory_path[64] = {0}; //TODO: Directory name could be up to 255 bytes by default
```
DeflateAwning, line 58: Well let's make this array 255 bytes long then? Or am I missing something?

firmware/Core/Src/telecommands/lfs_telecommand_defs.c
```diff
@@ -48,6 +47,82 @@ uint8_t TCMDEXEC_fs_unmount(const char *args_str, TCMD_TelecommandChannel_enum_t
...
+    if (parse_listing_count_result != 0) {
+        // error parsing
+        snprintf(
+            response_output_buf,
+            response_output_buf_len,
+            "Error parsing count arg: Error %d", parse_listing_count_result);
+        return 1;
+    }
+
+    int8_t list_directory_result = LFS_list_directory(arg_root_directory_path, (uint16_t) arg_listing_offset, (int16_t) arg_listing_count);
```
DeflateAwning, line 91: Use const

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -108,126 +102,151 @@ int8_t LFS_mount()
...
-    // result is positive on success, 0 at the end of directory, or  negative on failure.
+    if (count == 0) {
+        count = -1;
+    }
+
+    // result is positive on success, 0 at the end of directory, or negative on failure.
     int8_t read_dir_result = 1;
-    while (read_dir_result >= 0)
+    struct lfs_info info;
+    LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_NORMAL, LOG_all_sinks_except(LOG_SINK_FILE), "Name \t\t Bytes");
```
DeflateAwning, line 154: Can we please avoid using the `\t` tab character. It works funky, especially in the middle of lines.

If we put the size first, then alignment is much easier.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -108,126 +102,151 @@ int8_t LFS_mount()
...
-    return 0;
+    if (read_dir_result < 0) {
+        return read_dir_result;
+    } else {
+        LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_NORMAL, LOG_all_sinks_except(LOG_SINK_FILE), "Successfully listed contents from directory: %s", root_directory);
+        return 0;
+    }
 }
 
 /**
```
DeflateAwning, line 199: Please format docstrings with `///` instead of `/* */`

DeflateAwning (changes requested):
Just a couple tiny change suggestions, very nice otherwise though!

Sorry for the very slow review

firmware/Core/Src/telecommands/lfs_telecommand_defs.c
Saksham-P: I kept it this low hoping that in the future we could probably introduce metadata and other information directly in the file name and so the actual `file_name` would have a soft lock of 64 bytes. Although now that I am thinking about it, there really is no point for that. 

Do let me know what you think about it.

firmware/Core/Src/littlefs/littlefs_helper.c
Saksham-P: Do you think it would be a good idea to store all the directories and files in a buffer before printing them? This way we would be able to sort them using a certain criteria and also be able mimic `\t` by just using whitespace.

firmware/Core/Src/telecommands/lfs_telecommand_defs.c
DeflateAwning: Metadata (e.g., timestamps) make sense in the filename, yeah. 

The character array must be large enough to hold the longest allowable filename. Seems to me that the size of that array should just be bigger?

firmware/Core/Src/littlefs/littlefs_helper.c
DeflateAwning: >Do you think it would be a good idea to store all the directories and files in a buffer before printing them?

Probably not necessary, but potentially. I'm not too sure.

DeflateAwning:
I tacked on a small commit.

As long as you're okay with it, the output format is now like this (no more tabs):

```
========================= UART Telecommand Received =========================
CTS1+fs_list_directory(/,0,5)@tssent=1738997054050!
=========================
Parsed telecommand (len=51): 'CTS1+fs_list_directory(/,0,5)@tssent=1738997054050!'
0000000000000+0000011267_N [TELECOMMAND:NORMAL]: Executing telecommand from agenda slot 0, sent at tssent=1738997054050, scheduled for tsexec=0.
========================= Executing telecommand 'fs_list_directory'=========================
0000000000000+0000011411_N [LFS:NORMAL]: Name (Bytes)
0000000000000+0000011417_N [LFS:NORMAL]: ./
0000000000000+0000011421_N [LFS:NORMAL]: ../
0000000000000+0000011575_N [LFS:NORMAL]: demo_test.txt (37 B)
0000000000000+0000011598_N [LFS:NORMAL]: test_dir/
0000000000000+0000011603_N [LFS:NORMAL]: Successfully listed contents from directory: /
========================= Response (duration=322ms, err=0) =========================
```

DeflateAwning (approved):

PR #223, opened 2024-10-28
Implement SHA256 hash function

### LINKED ISSUE #38: Integrate a SHA256 hash function
Create a function called `GEN_sha256_hash(input_arr, input_arr_len, output_arr)`, where output_arr is 32 bytes long.

Compute the hash, using an existing implementation.

Create unit tests for short inputs, long inputs, inputs with nulls, and at least 20 random inputs generated with python (or similar).

This will be used in the telecommand uplink validation and authentication mechanism, as well as maybe others.

Benchmark the performance, and leave a comment on how long this takes to compute.

Makiv1:
resolves #38 

* Implements sha256 hash function which hashes a full message
* includes support for hashing chunks at a time instead of all in one go
* 20 random 4 killobyte alpha-numeric messages have been generated along with their expected hash and used to create 20 unit tests
* These unit tests are benchmarked when ran and the results logged
* Hashing 20 4 kilobyte messages takes ~490ms

sha256 implementation was taken from: https://github.com/B-Con/crypto-algorithms


DeflateAwning:
This is great! I'll just tack on a commit fixing a couple tiny things:

Notes/changes:
1. Added link to the source repo at the top of the files.
2. Use `///` docstrings instead of `/* */` docstrings.
3. Autoformat Python script (test generator) with Ruff.
4. Change Python script to generate `static const` variables (so that including that header file in more than one place doesn't break things).
5. In Python, use f-strings.
6. Add some `const` keywords (esp. on the input arg to the `CRYPT_compute_sha256_hash` function).
7. Add your benchmark comment.

DeflateAwning (approved):
Please see my comments above (describing the changes I made here)! If you're happy with these changes, feel free to merge!

DeflateAwning (approved):

DeflateAwning (approved):

DeflateAwning (approved):

Makiv1:
tests appear to be working correctly!

PR #229, opened 2024-11-02
Implement `TCMDEXEC_comms_dipole_switch_set_state`

### LINKED ISSUE #226: OBC Bring up Testing: Create Telecommand to Set Dipole Switch State
The UHF dipole switch allows the satellite to switch between what antenna is being used (out of the two options available). This is controlled by a single control signal from the STM, which is sent to pin “A”, while an inverter is used to send the inverted signal to pin “B”. Depending on the state at the A and B inputs (High/Low or Low/High), the respective antennas will be enabled (through RF1 or RF2).

A telecommand needs to be designed to test if the control signal to the switch is being sent correctly. The requirements are as follows.

- The control signal for the dipole switch comes out of the STM from pin 73 (PB12). 
- The telecommand should only take in one argument: Either 'A' or 'B'. 
- Based on the provided argument, the Telecommand should either pull PB12 to a HIGH state (if 'A') or a LOW state (if 'B'). 
- The state of this output should than be outputted through the Debug UART through a clear statement as a confirmation.
- Please name this telecommand "comms_dipole_switch_set_state"

DeflateAwning:
Fixes #226.

### DIFF
--- a/firmware/Core/Inc/comms_drivers/comms_drivers.h
+++ b/firmware/Core/Inc/comms_drivers/comms_drivers.h
@@ -1,0 +1,10 @@
+#ifndef INCLUDE_GUARD_COMMS_DRIVERS_H
+#define INCLUDE_GUARD_COMMS_DRIVERS_H
+
+#include <stdint.h>
+
+extern uint8_t COMMS_current_active_antenna;
+
+void COMMS_set_dipole_switch_state(uint8_t dipole_switch_antenna_num);
+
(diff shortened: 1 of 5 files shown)


NuclearTea (commented):
Looks good, just need to test it.

DeflateAwning:
@aaryandeshpande used this in testing, and all worked as expected.

NuclearTea (approved):
Code looks good and has been tested. 

Nice!

PR #231, opened 2024-11-08
Issue 53: Unique send id. stopped duplicate telecommand calls

Freddydeddydoo:
Added circular buffer to catch the last 100 telecommands.

Will add some unit tests to this but I have tested it!


NuclearTea:
@Freddydeddydoo Hey, just wondering if you have gotten to implementing those unit tests

Freddydeddydoo:
# Tests Steps : 

1. Run any telecommand with a timestamp of your choosing:  (ex: CTS1+hello_world()@tssent=5!).
2. Run that telecommand again with the same timestamp. 
3. check to see if it was parsed. The correct output would be :  
{
                    Telecommand skipped due to timestamp collision,
                    Timestamp sent: 5,
                    latest timestamp sent: 5
}
4. check to make sure that the telecommand didn't run afterwards.

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -53,6 +64,23 @@ uint8_t TCMD_add_tcmd_to_agenda(const TCMD_parsed_tcmd_to_execute_t *parsed_tcmd
...
 
+        // check to see if timestamp is in the circular buffer
+        for (uint32_t i = 0; i < TCMD_timestamp_sent_head; i++) {
+            if(parsed_tcmd->timestamp_sent == TCMD_timestamp_sent_buffer[i]) {
+                // Skip this telecommand
+                LOG_message(
+                    LOG_SYSTEM_TELECOMMAND, 
+                    LOG_SEVERITY_NORMAL, 
+                    LOG_SINK_ALL, 
+                    "Telecommand skipped due to timestamp collision\n"
```
DeflateAwning, line 75: Please remove `\n`.

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -43,8 +54,8 @@ const char* telecommand_channel_enum_to_str(TCMD_TelecommandChannel_enum_t chann
 /// @return 0 on success, 1 if the agenda is full.
 /// @note Performs a deep copy of the `parsed_tcmd` arg into the agenda.
 uint8_t TCMD_add_tcmd_to_agenda(const TCMD_parsed_tcmd_to_execute_t *parsed_tcmd) {
-    // FIXME: Issue #53 - Check `parsed_tcmd.timestamp_sent` against the yet-to-be-created `TCMD_timestamp_sent_record[]`, and skip this command if it is (to prevent replays).
-    // ^ Use a config variable to override this check, if desired.
+    // FIXME: If anyone wants please create unit tests for this. I am not sure if I should come back and add one but 
+    //we will see.
```
DeflateAwning, line 58: Please rephrase in the format of:

> TODO: Create unit test for <explain what the unit test is for>, if desired

I'd suggest that this might not be a place that a unit test makes a lot of sense

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -53,6 +64,23 @@ uint8_t TCMD_add_tcmd_to_agenda(const TCMD_parsed_tcmd_to_execute_t *parsed_tcmd
             continue;
         }
 
+        // check to see if timestamp is in the circular buffer
+        for (uint32_t i = 0; i < TCMD_timestamp_sent_head; i++) {
+            if(parsed_tcmd->timestamp_sent == TCMD_timestamp_sent_buffer[i]) {
+                // Skip this telecommand
+                LOG_message(
+                    LOG_SYSTEM_TELECOMMAND, 
+                    LOG_SEVERITY_NORMAL,
```
DeflateAwning, line 73: Should be a warning-level severity

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -17,6 +17,17 @@
...
+
+//the head of the circular buffer I think this should work
+
+//make these globals AND static.
+
+
+///@brief  The head of the circular buffer of timestamps of telecommands that have been sent.
+uint8_t  TCMD_timestamp_sent_head = 0;
+/// @brief The circular buffer of timestamps of telecommands that have been sent.
+uint64_t TCMD_timestamp_sent_buffer[TCMD_TIMESTAMP_RECORD_SIZE] = {0};
```
DeflateAwning, line 29: Consider renaming from "buffer" to "store" or "record". 

Buffers are typically temporary stores.

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -17,6 +17,17 @@
...
 uint64_t TCMD_latest_received_tcmd_timestamp_sent = 0;
 
+
+//the head of the circular buffer I think this should work
+
+//make these globals AND static.
+
+
+///@brief  The head of the circular buffer of timestamps of telecommands that have been sent.
+uint8_t  TCMD_timestamp_sent_head = 0;
```
DeflateAwning, line 27: Can we make this a u16? Can we also increase `TCMD_TIMESTAMP_RECORD_SIZE` to 500 (from 100)?

firmware/Core/Src/telecommands/telecommand_executor.c
```diff
@@ -17,6 +17,17 @@
 uint32_t TCMD_total_tcmd_queued_count = 0;
 uint64_t TCMD_latest_received_tcmd_timestamp_sent = 0;
 
+
+//the head of the circular buffer I think this should work
```
DeflateAwning, line 21: Inappropriate/unnecessary comment, I think

DeflateAwning (changes requested):
This looks awesome, great work! Exemplary circular buffer implementation

Just a couple of tiny fixes :)

Freddydeddydoo:
I have fixed these issues ill test it out again my computer USB drivers stopped working

NuclearTea:
I tested as well and seems to be working

NuclearTea:
@DeflateAwning I have tested and this seems to be working as expected. Just a quick question:
Since the tssent value is being used as unique id and currently, the circular buffer does not remove that value once that telecommand is executed. This means that you would have to send 500 telecommands until you could send any telecommand with the same tssent value. If this is the intended behaviour then this PR can be approved!

DeflateAwning:
I think that behaviour is good, yes.

Due to the forward-moving nature of time, I don't anticipate that we'll ever send a telecommand at an earlier timestamp than a previous telecommand was sent, thus making the "timestamp sent" a good unique identifier to deduplicate telecommands based on. 

A UUID would be an even better way to deduplicate telecommands, but UUIDs generally take more bytes to transmit, and don't carry as much valuable information. The tssent value can be used in a debugging manner, in addition to this duplicate execution lockout.

README.md
```diff
@@ -18,8 +18,9 @@ Please add your name to the list below in your first Pull Request!
 * Kale F.
 * Marko V.
 * Vaibhav K.
-* Muhammad Ali
+* Muhammad Ali.
```
DeflateAwning, line 21: I don't think "Ali" is short for anything.

DeflateAwning (approved):
Recommend fixing the README mistake, but looks good otherwise!

DeflateAwning:
Please make sure to delete your branches when you merge your PRs.

PR #233, opened 2024-11-09
Telecommand: LFS delete file (#227)

### LINKED ISSUE #227: Filesystem: Telecommand to delete a file

KaleF07:
Resolves #227:

**Summary**
- Included telecommand that uses `LFS_delete_file` helper function to delete a file from LFS
- Accepts 1 argument: The name of the file to be deleted

### DIFF
--- a/firmware/Core/Src/telecommands/lfs_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/lfs_telecommand_defs.c
@@ -180,7 +180,7 @@ uint8_t TCMDEXEC_fs_write_file_str(const char *args_str, TCMD_TelecommandChannel
 /// @brief Telecommand: Deletes a specified file in LittleFS
 /// @param args_str
 /// - Arg 0: File name to be deleted
-uint8_t TCMDEXEC_delete_file(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+uint8_t TCMDEXEC_fs_delete_file(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len) {
     char arg_file_name[64] = {0};
     const uint8_t parse_file_name_result = TCMD_extract_string_arg(args_str, 0, arg_file_name, sizeof(arg_file_name));
--- a/firmware/Core/Src/telecommands/lfs_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/lfs_telecommand_defs.c
@@ -183,7 +183,7 @@ uint8_t TCMDEXEC_fs_write_file_str(const char *args_str, TCMD_TelecommandChannel
 /// @note Do not add quotations around the argument, write as is.
 uint8_t TCMDEXEC_fs_delete_file(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len) {
-    char arg_file_name[64] = {0};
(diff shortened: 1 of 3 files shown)


NuclearTea (commented):
This all looks good. This just needs to be tested.

vaibhavkpr:
> This all looks good. This just needs to be tested.

I was able to quickly test this using my NAND memory module. Looks good!
✅ Able to successfully delete a file if it exists.
✅ Reports appropriate error when trying to delete a file that doesn't exist.

Here are some things to consider that I found during testing:
⚠️ I'm unable to reach the "Error parsing file name arg" scenario when I send an empty string "" as an input. Also to note as a result "" is accepted as a parameter when creating a file. Able to reach case: 3 (filename too long for parsing) successfully.
⚠️ Able to send input both with and without "", are both ways acceptable? May be worth including in the arg doc so it is more clear.

KaleF07:
> > This all looks good. This just needs to be tested.
> 
> I was able to quickly test this using my NAND memory module. Looks good! ✅ Able to successfully delete a file if it exists. ✅ Reports appropriate error when trying to delete a file that doesn't exist.
> 
> Here are some things to consider that I found during testing (while not directly linked to the issue itself but worth getting investigated if possible): ⚠️ I'm unable to reach the "Error parsing file name arg" scenario when I send an empty string "" as an input. Also to note as a result "" is accepted as a parameter when creating a file. Able to reach case: 3 (filename too long for parsing) successfully. ⚠️ Able to send input both with and without "", are both ways acceptable? May be worth including in the arg doc so it is more clear.

Thank you for testing! I'll take a look into the issues and get them fixed right away! Passing a filename argument of "" should not be acceptable, there should be an error case for an empty argument.

KaleF07:
While looking through the code, I found that the "Error parsing file name arg" condition was trivial since passing no arguments to this telecommand prompts a telecommand parsing error. However, I will keep the statement in the code just to be safe.

I added a note in the doc string indicating to not add quotations around the argument when parsing `fs_delete_file` to reduce any confusion.

KaleF07:
another good thing to mention is that it might be nice to wait until Saksham's `LFS telecommand list files [issue #7] ` is merged into main to merge this branch.

DeflateAwning:
PR #216 (list files in a directory) was just published. When you get a chance, would you mind rebasing and testing this? Then we can merge it!

firmware/Core/Src/telecommands/lfs_telecommand_defs.c
```diff
@@ -87,6 +88,33 @@ uint8_t TCMDEXEC_fs_write_file(const char *args_str, TCMD_TelecommandChannel_enu
     return 0;
 }
 
+/// @brief Telecommand: Deletes a specified file in LittleFS
+/// @param args_str
+/// - Arg 0: File name to be deleted
+/// @note Do not add quotations around the argument, write as is.
+uint8_t TCMDEXEC_fs_delete_file(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                        char *response_output_buf, uint16_t response_output_buf_len) {
+    char arg_file_name[64] = {0};
```
DeflateAwning, line 97: Please change `64` to `LFS_MAX_PATH_LENGTH` const (now in main).

DeflateAwning (changes requested):
Just that one change; looks good otherwise!

KaleF07:
All changes should be implemented, ready for a approval then we can merge!

DeflateAwning (approved):
Thanks!

PR #238, opened 2024-11-24
Added functionality to make warnings appear when LOG_message doesn't have same format as printf

2BlackCats:
Also fixed the warning that came up when functionality was added.

### DIFF
--- a/firmware/Core/Inc/log/log.h
+++ b/firmware/Core/Inc/log/log.h
@@ -49,7 +49,7 @@ enum {
     LOG_SYSTEM_ON = 1,
 };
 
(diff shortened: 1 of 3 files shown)


DeflateAwning:
This is really neat, nice! Would you mind explaining how this works and/or providing a link to where you learned this from?

DeflateAwning (approved):
I just tacked on two tiny commits that explain the change, and that set the size of the logged variable explicitly. In general, we prefer types like `uint32_t` instead of `int`/`long int`/etc.

Feel free to squash and merge this PR! If you haven't yet, feel free to add your name to the contributors in the README.

PR #242, opened 2024-11-30
Antenna deploy startup sequence

### LINKED ISSUE #148: Antenna Deploy: create sequence to call antenna deployment on initial system boot (post-deployment)
* Read a file from the flash memory (`lifecycle/deploy_antenna_on_boot_enabled.bool`). If it doesn't exist, write it as "true" (`1`). If it does exist, read it and see if it contains string `0` or `1`.
* If the flag file contains `1`, then check if the EPS is sufficiently charged.
* If the EPS is sufficiently charged, then try to deploy a random antenna (1 to 4).
* Loop.

Also, create a telecommand called `ant_set_deploy_antenna_on_boot_enabled`, which takes a 0 or 1 argument, and sets the file appropriately. Requires strong validation. Document that this telecommand shall be executed when the antenna is determined to be successfully deployed, as detemined by ground contact.

To deploy the antenna, use the `ANT_CMD_arm_antenna_system()` function, and then the `ANT_CMD_start_automated_sequential_deployment()` functions.

Makiv1:
resolves #148

At the moment there is lots of logging messages. Perhaps too many...

### DIFF
--- a/firmware/Core/Inc/rtos_tasks/rtos_tasks.h
+++ b/firmware/Core/Inc/rtos_tasks/rtos_tasks.h
@@ -14,5 +14,6 @@ void TASK_handle_uart_telecommands(void *argument);
 void TASK_execute_telecommands(void *argument);
 
 void TASK_monitor_freertos_memory(void *argument);
+void TASK_bootup(void *argument);
 
 #endif // __INCLUDE_GUARD__RTOS_TASKS_H__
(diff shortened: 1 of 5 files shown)


DeflateAwning:
Would you mind explaining what edge cases this covers? E.g., corrupted file system (and how it handles it), antenna fails to deploy (and how it handles it), battery low (...), etc. 

Just want to make sure this gets a well-thought-out review and test plan to accompany it.

Makiv1:
Whoopsies, I see now that I was not as thorough I a should have been... I'll make sure that we are handling all of these in the pr on Thursday.

DeflateAwning:
Ah sweet, thanks! I'm sure some of these are covered already; just wanted to understand what was here before I started reviewing the code :)

NuclearTea:
@Makiv1 Any luck with this issue?

Makiv1:
> @Makiv1 Any luck with this issue?
Very close to being done, still need to do a few tests with the eps before it's ready. Hoping to do those tomorrow!

Makiv1:
Closing this pr since I ended up messing up my git commits, opened a new branch and pr.

Issue #186, opened 2024-09-20
EPS: periodic logging of power usage, charging, etc.


DeflateAwning:
* Log EPS power usage, charging, etc. (i.e., housekeeping data) in a function.

### Power System Monitoring Rules
* Create configuration for each channel, with maximum current draw.
* If the current draw exceeds the associated maximum threshold, turn off the channel/system.

### Original Issue (Archived) - Missing Details
* Create an RTOS thread for this periodic logging.
* Power monitoring of subsystems when active; turn them off if they're having issues.

PR #259, opened 2025-01-17
EPS: periodic logging of power usage and disabling channels when issues are discovered.

2BlackCats:
Issue #186


firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -140,3 +142,84 @@ void TASK_time_sync(void *argument) {
...
+    
+    TASK_HELP_start_of_task();
+
+    uint32_t sleep_duration_ms = 30000; // 30s
+
+    osDelay(sleep_duration_ms);
+    sleep_duration_ms = 1000; // 1 second
+
+    EPS_struct_pdu_housekeeping_data_eng_t prev_EPS_pdu_housekeeping_data_eng;
+    EPS_CMD_get_pdu_housekeeping_data_eng(&prev_EPS_pdu_housekeeping_data_eng);
```
NuclearTea, line 156: This function can fail. Please get the response and ensure that its valid

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -140,3 +142,84 @@ void TASK_time_sync(void *argument) {
...
 }
+
+void TASK_EPS_power_monitoring(void *argument) {
+    
+    TASK_HELP_start_of_task();
+
+    uint32_t sleep_duration_ms = 30000; // 30s
+
+    osDelay(sleep_duration_ms);
+    sleep_duration_ms = 1000; // 1 second
```
NuclearTea, line 153: Rather than reassigning this, just use a new variable for the next use case. Not a big deal here, but in general, its better practice.

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -140,3 +142,84 @@ void TASK_time_sync(void *argument) {
...
+    EPS_vpid_eng_t prev_vpid_eng[32];
+    memcpy(prev_vpid_eng, prev_EPS_pdu_housekeeping_data_eng.vip_each_channel, sizeof(EPS_vpid_eng_t) * 32);
+
+    while (1) {
+        osDelay(sleep_duration_ms);
+        
+        char json_str[1000];
+        
+        EPS_struct_pdu_housekeeping_data_eng_t EPS_pdu_housekeeping_data_eng;
+        EPS_CMD_get_pdu_housekeeping_data_eng(&EPS_pdu_housekeeping_data_eng);
```
NuclearTea, line 166: Like stated above, this function can fail, ensure its valid

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -140,3 +142,84 @@ void TASK_time_sync(void *argument) {
...
+        EPS_vpid_eng_t vpid_eng[32];
+        memcpy(vpid_eng, EPS_pdu_housekeeping_data_eng.vip_each_channel, sizeof(EPS_vpid_eng_t) * 32);
+
+        //Power Logging
+        EPS_struct_pdu_housekeeping_data_eng_TO_json(&EPS_pdu_housekeeping_data_eng, json_str, 1000);
```
NuclearTea, line 171: This function can fail, ensure the return is valid

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -140,3 +142,84 @@ void TASK_time_sync(void *argument) {
...
+        );
+
+        uint32_t ch_bitfield = (EPS_pdu_housekeeping_data_eng.stat_ch_ext_on_bitfield << 16) & EPS_pdu_housekeeping_data_eng.stat_ch_on_bitfield;
+
+        //Power Monitoring
+        for (int channel = 0; channel < 32; channel++) {
+            if (ch_bitfield & 1) {
+
+                if (vpid_eng[channel].power_cW - prev_vpid_eng[channel].power_cW > 1000             //TODO: Check if the pdu values are correct
+                    || vpid_eng[channel].voltage_mV - prev_vpid_eng[channel].voltage_mV > 1000
```
NuclearTea, line 188: This and the above if statement can probably be combined. You could make 2 variables which represent each condition or just write all the statements together, either should be fine.

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -140,3 +142,84 @@ void TASK_time_sync(void *argument) {
...
+                    || vpid_eng[channel].current_mA - prev_vpid_eng[channel].current_mA > 1000
+                    || vpid_eng[channel].power_cW - prev_vpid_eng[channel].power_cW < -1000
+                    || vpid_eng[channel].voltage_mV - prev_vpid_eng[channel].voltage_mV < -1000
+                    || vpid_eng[channel].current_mA - prev_vpid_eng[channel].current_mA < -1000) {
+
+
+                    uint8_t result = EPS_CMD_output_bus_channel_off(channel);
+
+                    if (result != 0) {
```
NuclearTea, line 197: The Log_Message function in both the if and else block are essentially identical so you could use a ternary to output the appropriate message. Completely Optional as this is perfectly valid as well.

NuclearTea (commented):
Great Work overall.

Just some quality of life things.

Nice

DeflateAwning:
Can you please rebase? Can you please explain what criteria/thresholds you've selected for disabiling active channels.

Issues I see with the current approach:
1. Thresholds should be factored out into clearly-named variables.
2. I'm not sure this logic represents a failure mode at all: ` vpid_eng[channel].current_mA - prev_vpid_eng[channel].current_mA > 1000`. Why would the current draw changing instantaneously represent a fault condition? I think the actual way to implement this needs to be setting expected current draw maxima for each channel.
3. We should probably merge this with the EPS watchdog thread. We're getting too many threads, which will cause us to run out of memory. The actual actions parts should be refactored to a function, which is called from within a thread. Logging in one function, and power management in another function. 
4. Needs configuration variable for setting the period of the logging and which logs we want to enable. 

This looks like a solid start though, thanks!

2BlackCats:
@DeflateAwning made the thread a function and integrated it in to the watchdog. I didn't fully understand the "current draw maxima for each channel" and haven't changed that. If you could go over the code and give me your thoughts.

firmware/Core/Src/eps_drivers/eps_commands.c
```diff
@@ -6,6 +6,7 @@
 #include "eps_drivers/eps_types.h"
 #include "eps_drivers/eps_internal_drivers.h"
 #include "eps_drivers/eps_struct_packers.h"
+#include "eps_drivers/eps_power_management.h"
```
NuclearTea, line 9: Unsure why we are including this file and no other changes to the file

firmware/Core/Src/eps_drivers/eps_power_managment.c
```diff
@@ -0,0 +1,133 @@
+
+#include "eps_drivers/eps_power_management.h"
+
+uint32_t disableable_channels = 0x0001ffff;
```
NuclearTea, line 4: You are not using the configuration variables here. What you should do is initialize them in configuration.c and then include configuration.h in this file

firmware/Core/Src/eps_drivers/eps_power_managment.c
```diff
@@ -0,0 +1,133 @@
...
+    1000, 1000, 1000, 1000, 
+    1000, 1000, 1000, 1000, 
+    1000, 1000, 1000, 1000, 
+    1000, 1000, 1000, 1000};
+
+/**
+ * @brief Monitors the power consumption of each channel and logs the data in JSON format.
+ *
+ * @details This function will obtain the PDU housekeeping data and log the data in JSON format.
+ *          If this is the first time the function is called, it will save the data for future comparisons.
```
NuclearTea, line 19: Our doc strings are always ///, please change this

firmware/Core/Src/telecommands/eps_telecommands.c
```diff
@@ -520,3 +521,50 @@ uint8_t TCMDEXEC_eps_get_piu_housekeeping_data_run_avg_json(
     }
     return 0;
 }
+
+/// @brief Sets the EPS power managment threshold for the current of a specific channel.
+/// @param args_str 
+/// - Arg 0: The channel name or number (string).
+/// - Arg 1: Threshhold current (in mA) to set.
+/// @return 0 on success, >0 on failure
+/// @note Valid string values for Arg 0: "vbatt_stack", "stack_5v", "stack_3v3", "camera",
```
NuclearTea, line 530: If this is the case, please make a function that checks to ensure it

firmware/Core/Inc/config/configuration.h
```diff
@@ -20,6 +20,9 @@ static const uint8_t CONFIG_MAX_VARIABLE_NAME_LENGTH = 50;
 
 static const uint8_t CONFIG_MAX_JSON_STRING_LENGTH = UINT8_MAX;
 
+extern uint32_t disableable_channels;
```
NuclearTea, line 23: probably a better word than disableable

firmware/Core/Src/telecommands/eps_telecommands.c
```diff
@@ -520,3 +521,50 @@ uint8_t TCMDEXEC_eps_get_piu_housekeeping_data_run_avg_json(
...
+///     "uhf_antenna_deploy", "lora_module", "mpi", "boom".
+uint8_t TCMDEXEC_eps_power_managment_set_current_threshold(
```
NuclearTea, line 532: @DeflateAwning It seems because there is no way to create an int[] configuration variable, he has basically created another telecommand to update this. I think this is fine for now but we should open an issue to extend the configuration api to allow for int[]

firmware/Core/Src/config/configuration.c
```diff
@@ -31,6 +31,20 @@ CONFIG_integer_config_entry_t CONFIG_int_config_variables[] = {
         .variable_name = "TASK_heartbeat_period_ms",
         .num_config_var = &TASK_heartbeat_period_ms,
     },
+    //EPS Power Management Config
+    {
+        .variable_name = "disableable_channels",
```
NuclearTea, line 36: make sure to update this when you update the name

firmware/Core/Inc/eps_drivers/eps_power_management.h
```diff
@@ -0,0 +1,24 @@
...
+#include <string.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#include "rtos_tasks/rtos_eps_tasks.h"
+#include "eps_drivers/eps_types_to_json.h"
+#include "eps_drivers/eps_types.h"
+#include "eps_drivers/eps_commands.h"
+#include "log/log.h"
+#include "config/configuration.h"
```
NuclearTea, line 13: Only include files that you need in this one. If you .c file needs these, include this there

firmware/Core/Inc/rtos_tasks/rtos_eps_tasks.h
```diff
@@ -1,8 +1,6 @@
 #ifndef INCLUDE_GUARD_RTOS_EPS_TASKS_H__
 #define INCLUDE_GUARD_RTOS_EPS_TASKS_H__
 
-
-
```
NuclearTea, line 5: lets avoid removing whitespace just to reduce the chances of rebase errors

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -5,54 +5,87 @@
...
-/// see ISIS.EPS2.ICD.SW.IVID.7 pg. 85) seconds pass without a command being received. 
-/// The "peripheral reset" resets the communication peripheral on the eps's mcu 
-/// after 0.65 * TTC_WDG_TIMEOUT seconds. The "peripheral reset" allows for more graceful 
-/// resets if the eps's communication peripheral was responsible for loss of connection.
+uint32_t watchdog_timer = 7000, EPS_monitor_timer = 1000;       //TODO: Set watchdog timer and EPS monitor timer
+
+/**
+ * @brief Services the EPS watchdog periodically.
+ *
+ * The task services the watchdog periodically. If the task does not run for longer than
```
NuclearTea, line 22: use /// here as well

firmware/Core/Src/rtos_tasks/rtos_eps_tasks.c
```diff
@@ -5,54 +5,87 @@
...
     // Sleep 10s at the start so that more important tasks work first.
     // Important to service the watchdog near the start, though.
     osDelay(10000);
+    uint64_t last_time_watch_dog = TIM_get_current_unix_epoch_time_ms();
+
+    uint64_t last_time_EPS = TIM_get_current_unix_epoch_time_ms();
 
     while(1) {
-        const uint8_t result = EPS_CMD_watchdog();
+        if (TIM_get_current_unix_epoch_time_ms() - last_time_watch_dog > watchdog_timer) {
```
NuclearTea, line 47: Nice!

NuclearTea (commented):
Great Start!

Some quality of life changes.

2BlackCats:
@DeflateAwning How would I go about testing this considering that I would have to draw more current then set to toggle the channel off?

2BlackCats:
Current Test Plan:
Test 1: Connect board to eps and check response of eps.
Input required: 
->None
Expected Response: 
-> Log of all eps channels and no channels being disabled except those not connected.

Results: 

Test 2: Overload Current for a channel
Input: 
-> Cause current to overload channel 0 (VBATT)
Expected Response: 
->Log of all eps followed by a message for each channel (VBATT) that was disconnected about being disabled

Results:


Test 3: Test toggleable function (Disabled)
Input: 
-> CTS1+config_set_int_var(disableable_channels,0)! //Disable changing for all channels
-> CTS1+eps_set_channel_enabled(vbatt_stack,1)!
Expected Response:
-> No logged messages except for general eps log.

Results:

Test 4: Test disableable function (Enabled)
Input:
-> Continue from test 3
-> CTS1+config_set_int_var(disableable_channels,131071)!  // Enables for channel 0-16
Expected Response:
->Log message of disconnected channel (VBATT)

Results:

Test 5: Reconnect and turn on channel
Input:
-> Continue from test 4
-> draw current below threshhold
-> CTS1+eps_set_channel_enabled(vbatt_stack,1)!
Expected Response:
->Channel should not be disabled

Results:

NuclearTea:
@2BlackCats You need do the same thing as the other PR. Create a new branch from main and put in your changes. Your rebase has gone awry.

Issue #27, opened 2024-05-25
Create start-of-mission procedure to be run upon deployment from launch vehicle


DeflateAwning:
The start-of-mission execution flow should be:

1. Be deployed from launch vehicle. System powers on.
2. Start ~~60 minute~~ _30 minute_ timer, per ICD requirements.
3. Start beaconing through comms, once on each dipole switch.
4. Check EPS battery level. If <70%, wait until charged to 70%.
5. Attempt deploy comms antenna. Stop if battery <30%.
6. If antenna deploy fails, then retry (TBC how we determine this).
7. Attempt detumbling, while battery >30%.

DeflateAwning:
Related to #148

DeflateAwning:
Closed in favor of #420. This issue fails to utilize the RBF tag.

PR #260, opened 2025-01-17
Start up antenna deployment sequence

### LINKED ISSUE #148: Antenna Deploy: create sequence to call antenna deployment on initial system boot (post-deployment)

Makiv1:
Test plan is in the works. I will give a brief description of the antenna deployment procedure here:

1. check if deploy_antenna_on_boot_enabled.bool exists:
    * if yes check the value stored:
        * if 1 deploy antenna if eps is charged.
        * if 0 return
    * if no create file and write 1 into it, deploy antenna if eps is charged
2. If this fails 3 times (ie. 0 is not returned):
    * deploy antenna if eps is charged
3. if this fails another 2 times (5 failures in total):
    *deploy antenna

The thread sleeps for a given amount of time before retrying. The idea here is that if the file system fails or the eps is not correctly communicating its voltage, the antennas can still be deployed, since the program becomes less strict with its checks. Some tweaking is probably needed with the timeouts to ensure we are not ignoring our checks prematurely. 

Hopefully one of these weeks I'll get a chance to test it out with all of the actual hardware!

resolves #148 

(Work in progress) Test Plan: https://docs.google.com/document/d/1oRtDT4qSS5IVf1GBiwmX7NjGOQkH9OvaxN826IcmcE0/edit?tab=t.0

### DIFF
--- a/firmware/Core/Src/startup_sequence/antenna_deploy_startup.c
+++ b/firmware/Core/Src/startup_sequence/antenna_deploy_startup.c
@@ -4,6 +4,7 @@
 #include "log/log.h"
 #include "antenna_deploy_drivers/ant_commands.h"
 #include "antenna_deploy_drivers/ant_internal_drivers.h"
+#include "eps_commands.h"
 /// @brief Attempts to read to the file "lifecycle/deploy_antenna_on_boot_enabled.bool".
 /// If 1 is stored in the file then it attempts to deploy the antenna. If the file did not
 /// exist: then it creates the file, stores a 1 in it, and attempts to deploy.
@@ -31,7 +32,7 @@ static int16_t read_bool_from_deploy_antenna_on_boot_enabled_file (uint8_t* read
             return mount_result;
         }
     }
-    // lfs is mounted
+    // lfs is mounted at this point 
     
     lfs_file_t file;
     int16_t open_result = lfs_file_opencfg(&LFS_filesystem, &file, "lifecycle/deploy_antenna_on_boot_enabled.bool", LFS_O_RDONLY, &LFS_file_cfg);
@@ -55,6 +56,8 @@ static int16_t read_bool_from_deploy_antenna_on_boot_enabled_file (uint8_t* read
         );
         return open_result;
     }
+
+    // file exists and is opened for reading  
 
     int16_t read_result = lfs_file_read(&LFS_filesystem, &file, read_value, sizeof(uint8_t));
     if (read_result < 0) {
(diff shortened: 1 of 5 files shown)


Makiv1:
Hey I've started writing the test plan, I think it's worth checking out before reviewing since it describes the function of the code a bit better than I have done in the pr. Also discusses how the code might fail :)

https://docs.google.com/document/d/1oRtDT4qSS5IVf1GBiwmX7NjGOQkH9OvaxN826IcmcE0/edit?tab=t.0

firmware/Core/Src/startup_sequence/antenna_deploy_startup.c
```diff
@@ -0,0 +1,224 @@
...
+/// @brief Roughly computes a percentage of battery charge. If the battery is >50% charged, returns 1, else returns 0.
+/// @return 1 if the battery is >50% charged, 0 otherwise.
+static uint8_t eps_is_sufficiently_charged() {
+    //TODO: Not sure exactly which housekeeping data to use, assuming PBU?
+    EPS_struct_pbu_housekeeping_data_eng_t pbu_data;
+    if (EPS_CMD_get_pbu_housekeeping_data_eng(&pbu_data) != 0) {
+        return 0;
+    };
+    /* 
+    TODO: I have zero clue if this is correctly computing the battery pack voltage! please check Parker :)
```
NuclearTea, line 77: @DeflateAwning Please take a look at this. Also, shouldn't we have a function that calculates the current charge state of the different power units?

NuclearTea (commented):
Great Start!

Probably need some test plan to check this code

NuclearTea:
@Makiv1 The code for getting pbu battery percentage is merged

DeflateAwning:
Would love to test and merge this in the next week-ish. Very awesome stuff. Would you mind rebasing this and switching to use the new EPS battery percent function(s), and then we can test this!

DeflateAwning:
* Need to check the RBF pin state with a GPIO read.
* Need to do an "is antenna deployed" check and set a TX Inhibit variable for the AX100 (can't beacon until antennas deployed), I think

DeflateAwning:
Related to #27

DeflateAwning:
Going to close this in favour of the new plan in Issue #420

PR #270, opened 2025-02-14
Set up counter for GPS PPS

2BlackCats:
New PR for GPS pps counter for time keeping. Retested functionality still works. @NuclearTea


firmware/Core/Src/gpio_handler/gpio_handler.c
```diff
@@ -0,0 +1,19 @@
+
+#include "gpio_handler/gpio_handler.h"
+#include "log/log.h"
+#include "timekeeping/timekeeping.h"
+#include <stdio.h>
+#include <stdint.h>
+#include <string.h>
+uint32_t GPS_pps_counter = 0;
```
NuclearTea, line 8: If this is just going to be a 1 or 0, uint32 is too big I think

firmware/Core/Inc/stm32l4xx_hal_conf.h
```diff
@@ -182,7 +182,7 @@
   */
 
 #define  VDD_VALUE					  3300U /*!< Value of VDD in mv */
-#define  TICK_INT_PRIORITY            5U    /*!< tick interrupt priority */
+#define  TICK_INT_PRIORITY            15U    /*!< tick interrupt priority */
```
NuclearTea, line 185: Did you manually change the priority in the IOC?

NuclearTea (commented):
Great Work!

Couple of questions. Just some small changes needed

firmware/Core/Src/gpio_handler/gpio_handler.c
2BlackCats: @NuclearTea it is a counter to keep track of how often the system detects a pulse. So it wont just be a 1 or 0

firmware/Core/Inc/stm32l4xx_hal_conf.h
2BlackCats: No. I did it by it self. Should i change it back?

2BlackCats: @NuclearTea

2BlackCats:
Retested code. Doesn't receive multiple signals and will not accept signals too quickly.

NuclearTea (commented):
LGTM!

firmware/Core/Inc/gps/gps_pps_interrupt.h
```diff
@@ -0,0 +1,12 @@
+
+#ifndef __INCLUDE_GUARD__GPS_PPS_INTERRUPT_H
+#define __INCLUDE_GUARD__GPS_PPS_INTERRUPT_H
+
+#include "stm32l4xx_hal.h"
+#include "gpio_handler/gpio_handler.h"
+
+void GPIO_EXTI_Init(void);
```
NuclearTea, line 8: I was looking for this function in the STM32 low level drivers datasheet and I couldn't find it. I believe this is a function that you implemented? If so, can you rename this to GPS_PPS_EXTI_Init?

firmware/Core/Inc/gps/gps_pps_interrupt.h
```diff
@@ -0,0 +1,12 @@
...
+
+void Disable_GPIO_EXTI(void);
```
NuclearTea, line 10: Similar to above

firmware/Core/Src/gpio_handler/gpio_handler.c
```diff
@@ -0,0 +1,19 @@
+
+#include "gpio_handler/gpio_handler.h"
+#include "log/log.h"
+#include "timekeeping/timekeeping.h"
+#include <stdio.h>
```
NuclearTea, line 5: can you separate these to include sections:
```
#include "gpio.."
#include "log..."
#include "timekeeping"
<space>
#include <stdio..>
...
```

DeflateAwning:
Seems reasonable. IOC files will need to be made in a separate PR, and then reverted in this branch.

Would prefer to merge this when the full PPS functionality is actually implemented; in its current state, I'm not too sure this actually adds a distinctly beneficial feature to the satellite.

PR #272, opened 2025-02-19
Send arbitrary bytes to UART peripheral, and read its response

### LINKED ISSUE #19: Telecommand: send arbitrary bytes to UART peripheral, and read response
In case there's functionality of a UART peripheral we forgot to implement, we want to be able to send artibrary configuration commands, etc. to the peripheral (the MPI/GPS/Camera). 

### Args:
1. Which UART port to send on.
2. The bytes will be specified in ~~base64~~ _hex_.

DeflateAwning:
This is a rebase of https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/pull/185, with extra consideration taken for the GPS handler (which supports disabling the interrupt).

Fixes #19.

### DIFF
--- a/firmware/Core/Inc/log/log.h
+++ b/firmware/Core/Inc/log/log.h
@@ -3,6 +3,9 @@
 
 #include <stdint.h>
 
+// Messages up to 256 characters
+#define LOG_FORMATTED_MESSAGE_MAX_LENGTH 256
+
 typedef enum {
     LOG_SEVERITY_DEBUG = 1 << 0,
     LOG_SEVERITY_NORMAL = 1 << 1,
--- a/firmware/Core/Src/uart_handler/uart_handler.c
+++ b/firmware/Core/Src/uart_handler/uart_handler.c
@@ -7,7 +7,6 @@
 // Name the UART interfaces
 UART_HandleTypeDef *UART_telecommand_port_handle = &hlpuart1;
 UART_HandleTypeDef *UART_mpi_port_handle = &huart1;
-UART_HandleTypeDef *UART_lora_port_handle = &huart2;
 UART_HandleTypeDef *UART_gps_port_handle = &huart3;
 UART_HandleTypeDef *UART_camera_port_handle = &huart4;
 UART_HandleTypeDef *UART_eps_port_handle = &huart5;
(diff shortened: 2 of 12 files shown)


firmware/Core/Inc/log/log.h
```diff
@@ -3,6 +3,9 @@
 
 #include <stdint.h>
 
+// Messages up to 256 characters
+#define LOG_FORMATTED_MESSAGE_MAX_LENGTH 256
```
NuclearTea, line 7: The logs produced by the GPS often are larger than this, this was something that took me and Chris a while to figure out as we thought something was going wrong however, it was merely that the log message length was less than the gps response (ex: BESTXYZA returns ~265 bytes)

NuclearTea (commented):
LGTM.

Just a comment about log message length

NuclearTea:
testing steps for this:
https://docs.google.com/document/d/1RDm3sj04CtPXuz4nWWQtULcWHwBQGEti4S30fv1QXG8/edit?tab=t.0#heading=h.f1964er5owcq

NuclearTea:
# Testing Results for EPS (Connected to EPS): March 8

1. Testing with 00 as argument :heavy_check_mark: 
image

2. Testing getting the pbu housekeeping data eng json :heavy_check_mark: 
image

**Note: Validated that there are in fact, 97 bytes returned.**

3. Testing same as above, except the command code was changed to 0x60 rather than 0x62 :heavy_check_mark: 
image

DeflateAwning:
@NuclearTea Can you please give this a quick review and approve it if you're happy with it? Can continue testing as we use it with more hardware.

firmware/Core/Src/uart_handler/uart_handler.c
```diff
@@ -64,29 +93,122 @@ void HAL_UART_RxCpltCallback(UART_HandleTypeDef *huart) {
...
+        // Add the received byte to the buffer
+        UART_lora_buffer[UART_lora_buffer_write_idx++] = UART_lora_buffer_last_rx_byte;
+        UART_lora_last_write_time_ms = HAL_GetTick();
+
+        // Restart reception for next byte
+        HAL_UART_Receive_IT(UART_lora_port_handle, (uint8_t*) &UART_lora_buffer_last_rx_byte, 1);
+    }
+
+    // TODO: Implement function to utilize this DMA reception callback for the CAMERA
+    else if (huart->Instance == UART_camera_port_handle->Instance) {
```
NuclearTea, line 163: Not sure this makes sense as we are working on the camera stuff right now and it's implemented uisng DMA mode

NuclearTea (approved):
LGTM!

Just 2 things pop out to me.

My original comment about changing the log message length. This needs to be larger than 256, otherwise, when we try to log the GPS data, it will be cut off. I know this because Chris and I experienced this and were stuck for a couple of hours. It was simply due to the fact that a command like ```Bestxyza ...``` returns data that is more than 256 (around 265 ish).

Second is everything to do with the camera. I don't even know if this will work as the data coming from the camera is so large in quantity and quite fast that single byte based interrupt may not make sense. Additionally, we are currently implementing this using DMA mode. Having both doesn't make a lot of sense to me.

Would love to hear your thoughts.

Approved!

DeflateAwning:
Re the log message length - doesn't that just show up as two separate messages right now? Similar to how the EPS's 97 bytes (as hex) are split across 2 messages? If not, then can be adjusted. Increasing the log message length isn't a scalable solution (esp. not until we get confirmation on how large radio TX frames can be).

Re the camera port - I think there's a low chance this command-mode interaction will be used at all. Good to have it in case. You're right that it uses interrupt mode as-is though; guess that'd need to change to DMA (like the MPI is setup), darn.

Might make sense to just strip out the camera part completely for now.

DeflateAwning:
Just removed the UART2 LoRa UART stuff and pushed it into a separate backup branch in case we end up needing it as an extreme backup.

The camera only supports transmitting. Nice and simple. Camera will have to receive through the other DMA mechanism.

PR #274, opened 2025-02-21
Added log warning if time sync caused a change >2s. Added (untested) …

Ozone-Whiskey:
Need to test LOG messages with hardware (EPS & OBC clocks) or simulators. I need to sync eps to obc and vice versa to test logs and if they state which clock is being synced and which clock is being used as reference.


NuclearTea (commented):
Nice start!

I think you made  small mistake. Take a look please

firmware/Core/Src/telecommands/timekeeping_telecommand_defs.c
```diff
@@ -53,21 +56,48 @@ uint8_t TCMDEXEC_correct_system_time(const char *args_str, TCMD_TelecommandChann
...
 uint8_t TCMDEXEC_set_obc_time_based_on_eps_time(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len) {
+    EPS_struct_system_status_t status;
+
+    const uint8_t pre_result_status = EPS_CMD_get_system_status(&status);
+    if (pre_result_status != 0) {
+        return 1;
+    }
+
+    uint64_t obc_time_before_sync = ((uint64_t) status.unix_time_sec) * 1000;
```
NuclearTea, line 78: can you add the suffix `_ms` to the end. Do this for all variables that you declare in this function (have the variable name end with the unit of time,)

firmware/Core/Src/telecommands/timekeeping_telecommand_defs.c
```diff
@@ -53,21 +56,48 @@ uint8_t TCMDEXEC_correct_system_time(const char *args_str, TCMD_TelecommandChann
...
 }
 
 /// @brief Sync's obc time to eps time (+/- 1 second)
 /// @return 0 on success, >0 on failure.
 uint8_t TCMDEXEC_set_obc_time_based_on_eps_time(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
                         char *response_output_buf, uint16_t response_output_buf_len) {
+    EPS_struct_system_status_t status;
+
+    const uint8_t pre_result_status = EPS_CMD_get_system_status(&status);
+    if (pre_result_status != 0) {
```
NuclearTea, line 74: can you snprintf an error message here? Can you ensure that everywhere else in this function, when you return an error, that you also have an error message?

firmware/Core/Src/telecommands/timekeeping_telecommand_defs.c
```diff
@@ -53,21 +56,48 @@ uint8_t TCMDEXEC_correct_system_time(const char *args_str, TCMD_TelecommandChann
...
         return 1;
     }
-    snprintf(response_output_buf, response_output_buf_len,
-    "success syncing obc time");
+    snprintf(
+        response_output_buf, response_output_buf_len,
+        "success syncing obc time"
+    );
+
+    LOG_message(LOG_SYSTEM_ALL, LOG_SEVERITY_NORMAL, LOG_SINK_ALL, "CLock synced, OBC time before sync: %llu ms. OBC time after sync: %llu ms. Time difference: %llu \n", obc_time_before_sync, obc_time_after_sync, obc_time_after_sync-obc_time_before_sync);
```
NuclearTea, line 99: You are using ```%llu``` for the difference between the new time and the old time. This theoretically could be a negative value, however, the format specifier that you have used is for unsigned integers. Extract this out in the appropriate variable and use a format specifier that is for signed integers.

firmware/Core/Src/telecommands/timekeeping_telecommand_defs.c
```diff
@@ -76,7 +106,24 @@ uint8_t TCMDEXEC_set_obc_time_based_on_eps_time(const char *args_str, TCMD_Telec
 /// @return 0 on success, >0 on failure.
 uint8_t TCMDEXEC_set_eps_time_based_on_obc_time(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
```
NuclearTea, line 107: Similar comments for this function as for the above function

NuclearTea (commented):
Needs some small adjustments.

For the most part, it looks good!

firmware/Core/Src/telecommands/timekeeping_telecommand_defs.c
parker-research: Also, log messages should not end with `\n`.

Also, I don't think `%llu` works on our architecture. Unless I'm wrong, this finding suggests that this code has never been tested, which poses quite an extreme risk to this project. Code must be tested before being submitted for review.

Ozone-Whiskey: Gotcha. The PR had the awaiting Testing tag and mentions that it needed testing because I did not have access to OBC and EPS to test time changes.

parker-research: Ah fair enough, sorry, my bad

Ozone-Whiskey:
Tested telecommand sync time logs. Code seems ready. Need to test with EPS and OBC on Thursday.

NuclearTea:
@Ozone-Whiskey Did you rebase this branch or merge main into this branch? Either way, this PR now has 197 files changed. I think you need to revert the previous commit and do a rebase again. Something went wrong the last time it seems.

Issue #267, opened 2025-02-12
EPS: Create `EPS_convert_battery_voltage_to_percent` and `EPS_get_current_battery_percent` functions


DeflateAwning:
### `EPS_convert_battery_voltage_to_percent`

Linearly interpolate the voltage (argument in mV) to a percentage between 12V and 16V. 

Can go below 0% if the battery discharges below the 12V "minimum" safe threshold. Can go above 100% if the battery somehow charges above 16V.

Add unit tests. Does not rely on hardware.

### `EPS_get_current_battery_percent`

Fetch the appropriate housekeeping data (probably PBU [Power Battery Unit]). Then call `EPS_convert_battery_voltage_to_percent` on the relevant value, discarding the rest of the housekeeping data.

### Create a telecommand

Create a telecommand to demo this (named `eps_get_battery_percent`), which calls `EPS_get_current_battery_percent`

PR #275, opened 2025-02-22
TCMD for retrieving battery voltage percent from PBU

NuclearTea:
Completes #267.

TCMD for getting the battery voltage percent from PBU.

Independent function which calculates battery percentage with unit tests.

### DIFF
--- a/firmware/Core/Inc/eps_drivers/eps_calculations.h
+++ b/firmware/Core/Inc/eps_drivers/eps_calculations.h
@@ -1,0 +1,9 @@
+#ifndef INCLUDE_GUARD__EPS_CALCULATIONS
+#define INCLUDE_GUARD__EPS_CALCULATIONS
+
+#include "eps_drivers/eps_types.h"
+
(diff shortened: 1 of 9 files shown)


NuclearTea:
# Testing
## Success :heavy_check_mark: 

## Set up OBC with power supply voltage set to 12.73 V.
#### Output
image

## Setup OBC with power supply voltage set to 9.98 V
#### Output
image

## Setup OBC with power supply voltage set to 16.57 V
#### Output
image

DeflateAwning (approved):
Still requires testing with the actual EPS. Good enough for now though.

DeflateAwning (approved):
Thanks! Saw it during testing.

DeflateAwning:
Future thing to check - do we need to add that new linker flag to the GitHub Actions CI makefile? I forget exactly how I got it to work, and the build still passes, so not urgent

NuclearTea:
Testing March 8th, similar results as above, connected EPS to Actual OBC rather than using power supply

PR #278, opened 2025-03-01
Create int64_to_str function

DeemDeem52:
To fix a bug in the ADCS functions, I wrote this function. It's outside the ADCS scope, though, so here is a separate PR for it.

### DIFF
--- a/firmware/Core/Src/transforms/arrays.c
+++ b/firmware/Core/Src/transforms/arrays.c
@@ -2,7 +2,6 @@
 
 #include <ctype.h>
 #include <string.h>
-#include <stdlib.h>
 
 /// @brief Returns the index of the first character of the first occurrence of a substring in an array.
 /// @param haystack_arr The array to search within.
--- a/firmware/Core/Src/transforms/arrays.c
+++ b/firmware/Core/Src/transforms/arrays.c
@@ -111,11 +111,10 @@ void GEN_int64_to_str(int64_t value, char *buffer) {
 
     uint64_t absolute_value;
     
-    // Handle negative numbers
+    // Handle negative numbers (we can't use abs() for fixed-length integers)
     if (value < 0) {
         buffer[0] = '-';
-        // we can't use abs() for fixed-length integers
-        absolute_value = ~value + 1; 
+        absolute_value = -value; 
     } else {
         absolute_value = value;
     }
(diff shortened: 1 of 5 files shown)


firmware/Core/Src/transforms/arrays.c
```diff
@@ -100,6 +99,49 @@ void GEN_uint64_to_padded_str(uint64_t value, uint8_t padding_len, char *buffer)
...
+    if (value < 0) {
+        buffer[index++] = '-';
+        // we can't use abs() for fixed-length integers
+        absolute_value = ~value + 1; 
+    } else {
+        absolute_value = value;
+    }
+
+    // Convert the uint64_t value to a string in reverse order
+    while (absolute_value > 0) {
```
DeflateAwning, line 132: Could have used a call to the `uint64_to_str` function here. This is fine though.

firmware/Core/Src/unit_tests/test_byte_transforms.c
```diff
@@ -14,3 +15,37 @@ uint8_t TEST_EXEC__GEN_convert_big_endian_four_byte_array_to_uint32() {
...
     return 0;
 }
+
+uint8_t TEST_EXEC__GEN_int64_to_str() {
+    char buffer[32];
+    int64_t test;
+
+    test = 314159;
+    GEN_int64_to_str(test, &buffer[0]);
+    TEST_ASSERT_TRUE(buffer[0] == '3');
```
DeflateAwning, line 25: Suggestion: Use the strcmp function instead of checking each byte. Makes it a little clearer what's going on. 

This is fine; just not ideal.

DeflateAwning (dismissed):
2 suggests; up to you if you'd like to implement them. Good enough as-is.

DeemDeem52:
Thanks for your suggestions! I've implemented them.

firmware/Core/Src/transforms/arrays.c
```diff
@@ -100,6 +99,31 @@ void GEN_uint64_to_padded_str(uint64_t value, uint8_t padding_len, char *buffer)
...
+        return;
+    }
+
+    uint64_t absolute_value;
+    
+    // Handle negative numbers
+    if (value < 0) {
+        buffer[0] = '-';
+        // we can't use abs() for fixed-length integers
+        absolute_value = ~value + 1;
```
DeflateAwning, line 118: Hmm any reason this isn't just `absolute_value = -value;`

The `~value + 1;` construct is interesting to think about though

DeflateAwning (approved):
Just a review comment that might simplify it, but if you have a good reason, this is good too as-is

firmware/Core/Src/transforms/arrays.c
DeemDeem52: Somewhat embarrasingly, I hadn't thought of that.

DeflateAwning (approved):
:)

PR #280, opened 2025-03-09
LittleFS library upgraded to v2.10.1 (0494ce7)

Saksham-P:


DeflateAwning:
Looks like only a minor upgrade. Doesn't seem like any crazy changes. 

Might restrict upgrades after this one, unless there's a specific feature we want to use in a future upgrade.

DeflateAwning (approved):

PR #284, opened 2025-03-11
LittleFS Bug Fix: Data type conflicts causing overflows leading to unpredictable behavior 

Saksham-P:
Fixed bug where smaller size data types would cause returned function values to go outside type bounds:

For example:
`uint8_t temp1 = example_function();`

If the`example_function()` returns 257, then the `temp1` variable will have the value 1 because of overflow.


DeflateAwning (approved):
Looks good, thanks a lot!

PR #286, opened 2025-03-12
Fix integer width discrepency during substraction

NuclearTea:
- Fixes the failing unit test


DeflateAwning (dismissed):
Thanks! Didn't run it, so I'll trust you did.

Thanks for helping identify this @DeemDeem52

DeflateAwning (approved):
Just tacked on a docstring that mentions the perhaps-unexpected negative behaviour. Reapproved.

PR #292, opened 2025-03-15
Rename old I2C scanner to 'i2c_scan_bus_verbose', add smaller-output version

DeflateAwning:
Telecommand to test: `scan_i2c_bus`.

The old telecommand, which was previously named `scan_i2c_bus`, is now named `scan_i2c_bus_verbose`. Test with the flatsat using the ADCS.

### DIFF
--- a/firmware/Core/Src/telecommands/i2c_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/i2c_telecommand_defs.c
@@ -142,30 +142,38 @@ uint8_t TCMDEXEC_scan_i2c_bus(
         }
     }
 
-    // Print deviating addresses
     snprintf(
-        response_output_buf, response_output_buf_len, "Majority State: %s\n", 
+        response_output_buf, response_output_buf_len,
+        "{\"majority_state\":\"%s\"",
         majority_state == 0 ? "OK" : 
         majority_state == 1 ? "ERROR" : 
         majority_state == 2 ? "BUSY" : 
         majority_state == 3 ? "TIMEOUT" : "MISC"
     );
-
+    
     size_t buf_len = strlen(response_output_buf);
+    
     for (uint16_t i = 0; i < 128; i++) {
         if (address_states[i] != majority_state) {
             const size_t remaining_space = response_output_buf_len - buf_len - 1;
             if (remaining_space <= 0) break;
-
-            char msg[32];
-            snprintf(msg, sizeof(msg), "0x%02x: %s\n", i, 
-            address_states[i] == HAL_OK ? "OK" : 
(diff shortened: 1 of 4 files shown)


DeflateAwning:
Can someone please test this for me?

NuclearTea:
@DeflateAwning 2 questions. 

First, I think you deleted the original scan I2c function and pushed that into main, and here, you are putting it back in. Confused on whether you want that function or not.

Second, would it be sufficient to connect random i2c peripherals (like arduino components and such) to ensure that this is working properly? Are we looking for anything specific or just ensuring that the verbose prints out the correct output?

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
```diff
@@ -93,4 +94,78 @@ uint8_t TCMDEXEC_scan_i2c_bus(const char *args_str, TCMD_TelecommandChannel_enum
...
+/// @brief Scans the I2C bus for devices. Prints out the addresses of devices found.
+/// @param args_str
+/// - Arg 0: I2C bus to scan (1-4)
+/// @return 0 if successful, 1 if error.
+uint8_t TCMDEXEC_scan_i2c_bus(
+    const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+    char *response_output_buf, uint16_t response_output_buf_len
+) {
+    I2C_HandleTypeDef* hi2c;
+    const uint8_t bus_to_scan = atoi(args_str);
```
NuclearTea, line 108: I believe atoi returns an int which is a signed type of 4 bytes while you assign this variable an unsigned variable of 1 byte. Is this well defined behaviour, or are we assuming that the user will always pass in a valid value?

NuclearTea (commented):
Looks good!

firmware/Core/Src/telecommands/i2c_telecommand_defs.c
DeflateAwning: Should be fine for just this one as it's a debugging thing anyway; we'll be more careful in all other places though.

DeflateAwning:
Good catch on the confusion - fixed that with a rebase just now.

DeflateAwning:
Test successful (with 0, 1, and 2 responsive devices). Output format:

```
0946719292000+0000004273_E [TELECOMMAND:NORMAL]: 🚀 Executing telecommand 'scan_i2c_bus'.
0946719292000+0000004298_E [TELECOMMAND:NORMAL]: 🟢 Telecommand 'scan_i2c_bus' executed. Duration=17ms, err=0
{
  "majority_state": "ERROR",
  "0x20": "OK",
  "0x57": "OK"
}
```

PR #305, opened 2025-03-22
Create ADCS low power mode telecommand

### LINKED ISSUE #301: ADCS Telecommand: Set the ADCS into low power mode
This function will be called during the emergency low power transition state, but is also a good telecommand to have available.

DeemDeem52:
Just as soon as I know which components are essential for ADCS operation while the satellite is in low-power mode, this can be merged!

Closes #301


DeemDeem52:
`Power draw (Ch8) from PDU housekeeping data when ADCS is booted and powered, and no ADCS commands have been sent
    {
      "mV": 3349,
      "mA": 13,
      "cW": 11
    },

[note: within first 5 seconds of operation when ADCS is booting up, it draws ~40 mA]

When Channel 8 is off:
    {
      "mV": 333,
      "mA": -1,
      "cW": 0
    },


With enabled power (Run Mode On, all ADCS powers on **except** CubeStar, Motor, and GPS)
    {
      "mV": 3351,
      "mA": 35,
      "cW": 11
    },


Entering low power mode while keeping attitude stable (only Signal and Motor on)
    {
      "mV": 3352,
      "mA": 18,
      "cW": 10
    },

Entering low power mode without keeping attitude stable (none on)
    {
      "mV": 3351,
      "mA": 14,
      "cW": 8
    },

--> entering low power mode raises power to 44 mA immediately after as it processes the telecommand
    {
      "mV": 3350,
      "mA": 44,
      "cW": 7
    },

`

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -544,6 +544,57 @@ uint8_t TCMDEXEC_adcs_set_power_control(const char *args_str, TCMD_TelecommandCh
...
+    
+    uint64_t mode;
+    uint8_t status = TCMD_extract_uint64_arg(args_str, strlen(args_str), 0, &mode);
+    if (status != 0) {
+        snprintf(response_output_buf, response_output_buf_len,
+            "Telecommand argument extraction failed (err %d)", status);
+        return 1;
+    }
+
+    uint8_t attitude_mode = (uint8_t) mode;
```
DeflateAwning, line 562: Please mark as const.

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -544,6 +544,57 @@ uint8_t TCMDEXEC_adcs_set_power_control(const char *args_str, TCMD_TelecommandCh
...
 
+/// @brief Telecommand: Put the ADCS in low-power mode, with only essential component power.
+/// @param args_str 
+///     - Arg 0: whether to keep the attitude of the satellite stable (costs average 250 mW, maximum 1 W extra)
+/// @return 0 on success, >0 on error
+uint8_t TCMDEXEC_adcs_enter_low_power_mode(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+                                        char *response_output_buf, uint16_t response_output_buf_len) {
+    
+    uint64_t mode;
+    uint8_t status = TCMD_extract_uint64_arg(args_str, strlen(args_str), 0, &mode);
```
DeflateAwning, line 555: Please mark as const.

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -544,6 +544,57 @@ uint8_t TCMDEXEC_adcs_set_power_control(const char *args_str, TCMD_TelecommandCh
...
+    }
+
+    uint8_t attitude_mode = (uint8_t) mode;
+
+    if (attitude_mode > 1) {
+        snprintf(response_output_buf, response_output_buf_len,
+            "Stable-attitude mode must be 0 (off) or 1 (on); got %d", attitude_mode);
+        return 1;
+    } else if (attitude_mode) {
+        // in stable-attitude mode, we keep the CubeControl Signal and Motor power on if they were already on. If the satellite wasn't already stabilised, this will do nothing.
```
DeflateAwning, line 569: Please start sentences (and most comments in general) with capital letter.

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -544,6 +544,57 @@ uint8_t TCMDEXEC_adcs_set_power_control(const char *args_str, TCMD_TelecommandCh
...
+        // power cost: min 120 mW, max 200 mW
+        status = ADCS_set_power_control(ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF);
+    }
+    if (status != 0) {
+        snprintf(response_output_buf, response_output_buf_len, "Failed to disable ADCS peripherals"); 
+        return status;
+    }
+
+    // disable SD card logging
+    const uint8_t* temp_data_pointer[1] = {ADCS_SD_LOG_MASK_COMMUNICATION_STATUS};
```
DeflateAwning, line 583: Is this a pointer to an array (i.e., a double-pointer)? Why isn't this just an array?

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -544,6 +544,57 @@ uint8_t TCMDEXEC_adcs_set_power_control(const char *args_str, TCMD_TelecommandCh
     return status;
 }                            
 
+/// @brief Telecommand: Put the ADCS in low-power mode, with only essential component power.
+/// @param args_str 
+///     - Arg 0: whether to keep the attitude of the satellite stable (costs average 250 mW, maximum 1 W extra)
```
DeflateAwning, line 549: Please change to this (if accurate):

```
///     - Arg 0: Enable stable attitude mode. 1 to keep the attitude of the satellite stable (costs average 250 mW, maximum 1 W extra), 0 to disable control entirely.
```

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -544,6 +544,57 @@ uint8_t TCMDEXEC_adcs_set_power_control(const char *args_str, TCMD_TelecommandCh
...
+    uint8_t attitude_mode = (uint8_t) mode;
+
+    if (attitude_mode > 1) {
+        snprintf(response_output_buf, response_output_buf_len,
+            "Stable-attitude mode must be 0 (off) or 1 (on); got %d", attitude_mode);
+        return 1;
+    } else if (attitude_mode) {
+        // in stable-attitude mode, we keep the CubeControl Signal and Motor power on if they were already on. If the satellite wasn't already stabilised, this will do nothing.
+        // power cost: min 390 mW, average 470 mW, peak 1.05 W
+        status = ADCS_set_power_control(ADCS_POWER_SELECT_SAME, ADCS_POWER_SELECT_SAME, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF, ADCS_POWER_SELECT_OFF);
```
DeflateAwning, line 571: Can you please refactor this call out into a separate function, like `ADCS_enter_low_power_mode_with_stabilization` and `ADCS_enter_low_power_mode_without_stabilization` (or with the mode argument) so we can call that autonomously (i.e., so we can call the same function that this telecommand calls directly).

DeflateAwning (changes requested):
Looks good, thanks! Just a few changes.

DeflateAwning:
How do we exit this mode? Is there a "Ctrl+Z" equivalent, or do we need to run a bunch of telecommands?

DeemDeem52:
To exit low power mode, we just call ADCS_set_power_control with whatever peripherals we want to turn on (the ADCS doesn't really have a proper low-power mode, so it doesn't really change anything.)

firmware/Core/Src/telecommands/telecommand_adcs.c
DeemDeem52: It's a pointer to an array. The reason is for commissioning purposes, we want to be able to transmit a bunch of arrays ORed with each other. In this case we don't need that functionality, but we need to use the same type to make it a valid ADCS command.

firmware/Core/Src/telecommands/telecommand_adcs.c
DeemDeem52: I'm not sure what you mean -- we change this variable later when we are getting the statuses of the other ADCS commands.

DeflateAwning: Please use separate variables for each status request

DeflateAwning: Edit: While I generally prefer that approach, I'm seeing now that there's many points in the ADCS code where that makes less sense (e.g, when you call different ADCS commands in a switch statement). 

The general rationale behind this request/suggestion is that you miss out on "unused variable" warnings, and that it's tougher to follow in review what the status is from (esp. if `status` is only overridden conditionally).

firmware/Core/Src/telecommands/telecommand_adcs.c
DeflateAwning: Gotcha. After thinking about it more, this actually seems fine.

DeflateAwning:
### New Comment 1
After digging into the double-pointer code a bit more, I see that the number 10 (length of those arrays, loop upper bound, etc.) is repeated in a ton of places as a magic number. Can we factor that out to a `#define` to make it more clear? Perhaps like `ADCS_SD_LOG_BITFIELD_LENGTH_BYTES` (or something else if I'm misunderstanding what it is). E.g., `ADCS_SD_LOG_MASK_COMMUNICATION_STATUS[10]` -> `ADCS_SD_LOG_MASK_COMMUNICATION_STATUS[ADCS_SD_LOG_BITFIELD_LENGTH_BYTES]`

Can we then add it to the docstring of all SD-related functions that expect an array of pointer-to-arrays, where each array is `ADCS_SD_LOG_BITFIELD_LENGTH_BYTES` bytes long? E.g., `ADCS_combine_sd_log_bitmasks(...)` docstring.

### New Comment 2
In functions like `ADCS_set_sd_log_config`, can you please rename argument `log_array_size` to `log_array_len`. Size is generally measured in bytes (not elements), and refers to to maximum length (as opposed to the filled/readable length). Length (shortened to `_len`) is more appropriate for `ADCS_set_sd_log_config()`, `ADCS_combine_sd_log_bitmasks()`, etc.).

DeemDeem52:
Thanks Parker! I've made the changes you requested.

firmware/Core/Src/adcs_drivers/adcs_internal_drivers.c
```diff
@@ -198,12 +198,12 @@ uint8_t ADCS_convert_double_to_string(double input, uint8_t precision, char* out
...
 /// @param[in] array_in Array of pointers to 10-byte uint8 data arrays
-/// @param[in] array_in_size Size of the array_in array
+/// @param[in] array_in_len Size of the array_in array
 /// @param[out] array_out 10-byte uint8 data array to send the result to
 /// @return 0 once complete.
-uint8_t ADCS_combine_sd_log_bitmasks(const uint8_t **array_in, const uint8_t array_in_size, uint8_t *array_out) {
-    for (uint8_t i = 0; i < array_in_size; i++) {
-        for (uint8_t j = 0; j < 10; j++) {
+uint8_t ADCS_combine_sd_log_bitmasks(const uint8_t **array_in, const uint8_t array_in_len, uint8_t *array_out) {
+    for (uint8_t i = 0; i < array_in_len; i++) {
```
DeflateAwning, line 205: Logical but here: `array_out` could contain garbage, which is never zeroed. I think you should do a `memset` to set `array_out` to all zeros before starting the loops

DeflateAwning (changes requested):
Looks great, thanks! 

Just noticed one more thing scrolling through

DeflateAwning (approved):
Thanks

PR #308, opened 2025-03-24
Integrate camera init/test functions and telecommands

DeflateAwning:


DeflateAwning:
@MrBear101, once this is merged, make sure to base your code off of this.

@NuclearTea, would appreciate if you could try to review this in the next day or two so we can build on top of this (for the new camera code, and for the other self-check branch)

firmware/Core/Src/camera/camera_init.c
```diff
@@ -0,0 +1,179 @@
...
+            rx_status
+        );
+
+        return 3; // Error
+    }
+
+    // Safety: Ensure the string is null-terminated.
+    UART_test_str[sizeof(UART_test_str) - 1] = 0;
+
+    DEBUG_uart_print_str("Camera test response:\n");
```
NuclearTea, line 151: Unless you have a very good reason, I would suggest using the Log_message function here

firmware/Core/Src/camera/camera_init.c
```diff
@@ -0,0 +1,179 @@
...
+    DEBUG_uart_print_str((char *) UART_test_str);
+    DEBUG_uart_print_str("\n");
+    DEBUG_uart_print_str("END RESPONSE\n");
+
+    if (UART_test_str[0] == 0) {
+        UART_test_str[0] = 0;
```
NuclearTea, line 157: I think this is a typo lol

NuclearTea (commented):
Looks Good Overall.
Just some small things I noted.

Great Job!

firmware/Core/Src/camera/camera_init.c
DeflateAwning: Ha nice catch, thanks

firmware/Core/Src/camera/camera_init.c
DeflateAwning: Was afriad of the messages being too long, but agreed. Done.

DeflateAwning:
Tested like this. Is good.

PR #310, opened 2025-03-26
Config variable: reset after system uptime exceeds this many seconds

2BlackCats:
Waiting for PR #186 before merging.

### DIFF
--- a/firmware/Core/Inc/config/configuration.h
+++ b/firmware/Core/Inc/config/configuration.h
@@ -22,6 +22,7 @@ static const uint8_t CONFIG_MAX_JSON_STRING_LENGTH = UINT8_MAX;
 
 // extern for EPS
 extern uint32_t EPS_monitor_interval_ms;
+extern uint32_t system_reset_interval;
 
 // extern
 extern const uint32_t CONFIG_freertos_min_remaining_stack_percent;
--- a/firmware/Core/Inc/config/configuration.h
+++ b/firmware/Core/Inc/config/configuration.h
@@ -22,7 +22,7 @@ static const uint8_t CONFIG_MAX_JSON_STRING_LENGTH = UINT8_MAX;
 
 // extern for EPS
 extern uint32_t EPS_monitor_interval_ms;
-extern uint32_t system_reset_interval;
+extern uint32_t STM32_system_reset_interval_ms;
 
 // extern
 extern const uint32_t CONFIG_freertos_min_remaining_stack_percent;
--- a/firmware/Core/Src/config/configuration.c
+++ b/firmware/Core/Src/config/configuration.c
@@ -13,7 +13,7 @@ uint32_t CONFIG_int_demo_var_2 = 242344;
 
 // extern
 uint32_t EPS_monitor_interval_ms = 20000;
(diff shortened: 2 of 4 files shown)


NuclearTea (commented):
LGTM!

Waiting on that other PR to be merged

DeflateAwning:
Thanks for taking this quick one :)

Once the other one is merged (just reviewed it now), I'd recommend that you pull main, and then do an interactive rebase: `git rebase -i main`. You'll be prompted with a text file showing all commits. Carefully change the word "pick" to "d" (for drop or delete) on all the commits from BEFORE this PR specifically. That is, you will drop all the commits which are already merged into main in the other branch. Then, save-and-close that file, and this branch will be perfect.

I only explain this explicitly because I have a feeling you might encounter a few of these chained cases with as fast as you work :)

DeflateAwning:
Can you please rebase onto main now that the other one is merged?

firmware/Core/Inc/config/configuration.h
```diff
@@ -22,6 +22,7 @@ static const uint8_t CONFIG_MAX_JSON_STRING_LENGTH = UINT8_MAX;
 
 // extern for EPS
 extern uint32_t EPS_monitor_interval_ms;
+extern uint32_t system_reset_interval;
```
DeflateAwning, line 25: Please rename from `system_reset_interval` to `STM32_system_reset_interval_ms`.

All variables containing a numerical quantity require units. All variables require a namespace prefix.

firmware/Core/Src/config/configuration.c
```diff
@@ -13,6 +13,7 @@ uint32_t CONFIG_int_demo_var_2 = 242344;
 
 // extern
 uint32_t EPS_monitor_interval_ms = 20000;
+uint32_t system_reset_interval = 604800000;
```
DeflateAwning, line 16: The highest number a uint32 can store is about 50 days, so uint32 seems reasonable. 

Can you please leave a comment saying how long this default value is? Looks to me like 604800000ms = 7 days.

firmware/Core/Src/rtos_tasks/rtos_background_upkeep.c
```diff
@@ -40,6 +41,15 @@ void TASK_background_upkeep(void *argument) {
             EPS_monitor_last_uptime = current_time;
         }
         
+        if (start_time_before_reset + system_reset_interval < TIM_get_current_unix_epoch_time_ms()) {
```
DeflateAwning, line 44: Can you please just do a comparison of `if (TIM_get_current_system_uptime_ms() > system_reset_interval_ms) then reset` instead of bringing in the unix timestamps?

Unix timestamps and time syncs add complexity which isn't required/may be problematic (e.g., if the time is synced incorrectly, it could lead to some sort of reset loop here).

DeflateAwning (changes requested):
Looks awesome, thanks! Just a couple tweaks

DeflateAwning (approved):
Awesome, thank you! I just tacked on a quick commit which includes the reset time in the log message - may be helpful for if we ever decrease that timeout, and forget what we decreased it to.

Issue #248, opened 2024-12-02
GPS: ASCII parsers for select log command responses


christopherkinyua:
This ticket is for creating functions to parse the log responses for the following functions:
- BESTXYZA
- TIMEA

Unit tests for the above parsers and helping functions will be required

PR #311, opened 2025-03-26
GPS ASCII Parsers and Unit Test

christopherkinyua:
Summary
This PR contains functions to perform ASCII parsing for the following GPS log functions:

BESTXYZA
TIMEA
Respective unit tests have been created for the functions to achieve this.

Testing Instructions
I verified the accuracy of the GPS ASCII parsing functions by running the run_all_unit_tests telecommand and ensured that all my unit tests passed. I left my LOG_message function calls in to show the error being invoked. However, this will be removed during the review to declutter the unit test call function with all the prints to UART.


DeflateAwning:
New version of PR #276. Fixes #248.

DeflateAwning (approved):
Re-approved, thanks. Approving on the assumption it's about the same as the previous version.

Issue #253, opened 2024-12-13
GPS Telecommand: Enable/disable the GPS (low power mode)


DeflateAwning:
The GPS's power is always connected (and cannot be disabled from the EPS). We must implement commands to enable and disable it, and must ensure it is disabled by default on startup. 

### Requirements:
1. Function which sets the state of the GPS (investigation into states required).
2. Telecommand which calls that function
3. In a thread, especially at boot-up, the GPS must be disabled. Perhaps in the general start-up sequence, telling the GPS to turn off would be good. A thread which tells the GPS to be disabled every 30 seconds might also be valuable, in case the GPS experiences a fault, restarts itself, and turns back on.

PR #247, opened 2024-12-02
GPS: Send log command and receive log response from gps receiver

christopherkinyua:
## Related Issues

Closes #246  

## Summary

This PR implements UART communication between the GNSS Receiver and the OBC. This includes transmitting a log command as a string and receiving a log response from the GNSS receiver

## Changes Made

The changes made to realize this were as follows:
- Modified `uart_handler` to receive the GNSS response in interrupt mode. 
- Created a function `GPS_set_uart_interrupt_state` to initialize UART communication
- Created a file called `gps_internal_drivers` that contains the function, `GPS_send_cmd_get_response`, that handles the communication with the GNSS receiver.
- Added a telecommand `TCMDEXEC_gps_send_log_cmd` to call the function mentioned above.

## Testing Instructions
Tests were conducted on 28th November with @DeflateAwning and @NuclearTea present and we verified the functionality of the functions mentioned. Hence, the tests were successful. 

## Special Notes for Your Reviewer(s)

There are some `TODO` and `FIXME` tags within the code based on the current implementation that will need deeper looking into for proper functionality.

PR #316, opened 2025-04-04
GNSS: Create power control telecommand

celinahy:
**NOTE**: The logic to disable/enable TRACKSIGNAL has been removed because of ambiguity on whether we can use it/if it will work with our antenna model. Notes have been kept in for posterity's sake.

This is related to issue [#253](https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/issues/253).

---

## 🔧 Changes Introduced in This PR

- Added to the file `gps_internal_drivers` to house the primary function that will either disable or enable a single function of the GPS.

- Added ~~two~~ one telecommands:
  ~~1. **`TCMDEXEC_gps_enable_disable_command`**  
     Allows you to either disable or enable a single function of the GPS.~~

  2. **`TCMDEXEC_gps_enable_disable_all`**  
     Allows you to either disable or enable all possible functions of the GPS for "low power" mode, which currently is only ANTENNAPOWER.

---

## ✅ Testing

Testing was conducted on **April 4th, 2025** with **@NuclearTea**

### 1. Testing `ANTENNAPOWER` command
- Verified that disabling and enabling `ANTENNAPOWER` through `TCMDEXEC_gps_enable_disable_command` worked as expected.

**Findings:**

| State         | Channel 5 Power Draw |
|---------------|----------------------|
| Enabled       | ~510 mA               |
| Disabled      | ~476 mA               |

---

### 2. ~~Testing continuous enable/disable using `TCMDEXEC_gps_enable_disable_all`~~ NO LONGER APPLICABLE TO THIS ISSUE/PR
- Populated `TCMD_GPS_COMMAND_NAMES` with `'ANTENNAPOWER'` **23 times** to simulate:
  - 1 command for `ANTENNAPOWER`
  - 22 commands for `TRACKSIGNAL` (for 22 signals)

- Found that executing the telecommand started off normally, but a **buffer overflow or similar issue** caused a **software reboot**.

Buffer Overflow Screenshot

---

### 3. ~~Measuring `TRACKSIGNAL` Power Impact via `TCMDEXEC_gps_send_log_cmd`~~ NO LONGER APPLICABLE TO THIS ISSUE/PR
Even though `TRACKSIGNAL` logic hasn’t been added to the new telecommands yet, we still measured its power impact using the already-tested `TCMDEXEC_gps_send_log_cmd` from this merged [PR](https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/pull/247).

**Issues Encountered:**
- `ERROR: Invalid Message ID`
- `ERROR: Not allowed for input`

These will be resolved after further communication with NovAtel.

Invalid Message ID
Not Allowed for Input

### DIFF
--- a/firmware/Core/Src/gps/gps_internal_drivers.c
+++ b/firmware/Core/Src/gps/gps_internal_drivers.c
@@ -168,7 +168,7 @@ uint8_t GPS_set_power_enabled(uint16_t enable_disable_flag)
         
         // Send log command to GPS and receive response
         const uint8_t gps_cmd_response = GPS_send_cmd_get_response(
-            full_command, full_command_len, GPS_rx_buffer, GPS_rx_buffer_len, GPS_rx_buffer_max_size);
+            full_command, full_command_len, GPS_rx_buffer, GPS_rx_buffer_max_size, &GPS_rx_buffer_len);
 
         if (gps_cmd_response != 0)
         {
--- a/firmware/Core/Src/gps/gps_internal_drivers.c
+++ b/firmware/Core/Src/gps/gps_internal_drivers.c
@@ -149,7 +149,7 @@ uint8_t GPS_send_cmd_get_response(
 /// @param rx_buf_len Length of the response buffer.
 /// @param rx_buf_max_size Maximum length of the response buffer.
 /// @return 0 on success, >0 if error.
-uint8_t GPS_enable_disable(const char *cmd_arg, uint16_t enable_disable_flag, uint8_t rx_buf[], uint16_t rx_buf_len, const uint16_t rx_buf_max_size)
+uint8_t GPS_set_power_enabled(const char *cmd_arg, uint16_t enable_disable_flag, uint8_t rx_buf[], uint16_t rx_buf_len, const uint16_t rx_buf_max_size)
 {
     // Extract Arg 0: incomplete command to send
     char full_command[128];
--- a/firmware/Core/Src/gps/gps_internal_drivers.c
+++ b/firmware/Core/Src/gps/gps_internal_drivers.c
@@ -156,14 +156,6 @@ uint8_t GPS_enable_disable(const char *cmd_arg, uint16_t enable_disable_flag, ui
     snprintf(full_command, sizeof(full_command), "%s %s\n", cmd_arg, enable_disable_flag == 1 ? "ON" : "OFF");
     const uint16_t full_command_len = strlen(full_command);
 
-    // Testing purposes to be removed once out of Draft PR state
-    LOG_message(
-        LOG_SYSTEM_GPS,
-        LOG_SEVERITY_NORMAL,
-        LOG_SINK_ALL,
-        "THIS IS THE FULL COMMAND TO BE SENT TO OEM: %s",
-        full_command);
-
     // Send log command to GPS and receive response
     const uint8_t gps_cmd_response = GPS_send_cmd_get_response(
         full_command, full_command_len, rx_buf, rx_buf_len, rx_buf_max_size);
--- a/firmware/Core/Src/gps/gps_internal_drivers.c
+++ b/firmware/Core/Src/gps/gps_internal_drivers.c
@@ -16,6 +16,9 @@ extern UART_HandleTypeDef *UART_gps_port_handle;
 const uint32_t GPS_RX_TIMEOUT_BEFORE_FIRST_BYTE_MS = 800;
 const uint32_t GPS_RX_TIMEOUT_BETWEEN_BYTES_MS = 800;
 
+const char *GPS_LOW_POWER_COMMANDS[] = { "ANTENNAPOWER" };
+const uint8_t GPS_LOW_POWER_COMMANDS_LEN = 1;
+
 /// @brief Sends a log command to the GPS, and receives the response.
 /// @param cmd_buf log command string to send to the GPS.
 /// @param cmd_buf_len Exact length of the log command string.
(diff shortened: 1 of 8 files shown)


firmware/Core/Src/gps/gps_internal_drivers.c
```diff
@@ -133,3 +133,45 @@ uint8_t GPS_send_cmd_get_response(const char *cmd_buf, uint8_t cmd_buf_len, uint
...
+/// @return 0 on success, >0 if error.
+uint8_t GPS_enable_disable(const char *cmd_arg, uint16_t enable_disable_flag, uint8_t rx_buf[], uint16_t rx_buf_len, const uint16_t rx_buf_max_size)
+{
+    // Extract Arg 0: incomplete command to send
+    char full_command[128];
+    snprintf(full_command, sizeof(full_command), "%s %s\n", cmd_arg, enable_disable_flag == 1 ? "ON" : "OFF");
+    const uint16_t full_command_len = strlen(full_command);
+
+    // Testing purposes to be removed once out of Draft PR state
+    LOG_message(
```
NuclearTea, line 152: Please Remove this

firmware/Core/Src/telecommands/gps_telecommand_defs.c
NuclearTea, line 1: For this entire file, can you remove the whitespace changes? They can cause rebase issues sometimes lol

NuclearTea (commented):
Overall Looks good!

just some small things to clean up

firmware/Core/Src/gps/gps_internal_drivers.c
```diff
@@ -133,3 +133,45 @@ uint8_t GPS_send_cmd_get_response(const char *cmd_buf, uint8_t cmd_buf_len, uint
...
 }
+
+/// @brief Sends an enable/disable command to the GPS, and receives the response.
+/// @param cmd_arg the incomplete command (eg: ANTENNAPOWER)
+/// @param enable_disable_flag 1 to enable (power on), 0 to disable (power off)
+/// @param rx_buf Buffer to store the response.
+/// @param rx_buf_len Length of the response buffer.
+/// @param rx_buf_max_size Maximum length of the response buffer.
+/// @return 0 on success, >0 if error.
+uint8_t GPS_enable_disable(const char *cmd_arg, uint16_t enable_disable_flag, uint8_t rx_buf[], uint16_t rx_buf_len, const uint16_t rx_buf_max_size)
```
DeflateAwning, line 144: 1. Rename function to `GPS_set_power_enabled`.
2. Please significantly simplify the function interface (i.e., remove arguments). I think the only argument should be `uint8_t enabled`.
3. The `cmd_arg` values should be contained within this function, and should iterate through all of them. 

The point of this function is to have a one-stop command to enter into low-power mode. The functionality implemented here (i.e., a telecommand which sets the power state of a single arbitrary GNSS subsystem) already exists with the normal "execute any command on the GNSS" command.

Looking in the other parts of the code, it looks like the only part we're changing the power of is `"ANTENNAPOWER"`. Are there no other options we can disable power for?

DeflateAwning (changes requested):

DeflateAwning:
We need to monitor just the power of the GNSS when it's connected in the lab with the antenna. Seeing the decrease with the EPS is great, but more important is validating the absolute current draw. EPS Channel 5 has too much other stuff on it to assess only the GNSS current draw.

firmware/Core/Src/telecommands/gps_telecommand_defs.c
celinahy: yuppers. done. do we have a standardized formatter extension? Sometimes when I mess up tabs, by habit I do shift + option + f, which default formats. It might be the fallback formatter provided by VSCode, as I can't find anywhere that i've installed an extension.

firmware/Core/Src/gps/gps_internal_drivers.c
celinahy: 1) will do!
2) Do we care about buffers? Can we keep the args `uint8_t rx_buf[]`, `uint16_t rx_buf_len`, and `const uint16_t rx_buf_max_size` in the function interface? By passing in these vars, there is already a buffer that can hold the response from calling `GPS_send_cmd_get_response` instead of creating a new buffer for this. 
3) I agree, keeping the list of functions to call, at the lowest level possibly is ideal. 

Gotcha. Rest assured there was a lot of collaboration with Orbit (Aarti and Meagan) on what we can and cannot disable on the OEM7.

 I first gathered a large list of commands from the OEM7 manual that can be enabled and disabled, and went through the gruelling process with Orbit to discuss and further investigate if each command is applicable and safe for our use case. Some functions were a hard 'don't touch', others were disabled by default, and any outstanding questions were directed to NovAtel for further guidance. This is all documented under the `OEM7 commands for disable/enable function` file under the Software folder on our team drive. I also believe Aarti sent the final list to them to confirm if it was a good idea to enable/disable. For better or for worse, there were only two commands that were deemed safe to touch for this low power mode issue, `ANTENNAPOWER` and `TRACKSIGNAL`(the latter which we couldn't even test that one day, and we're waiting back from NovAtel for what they have to say about that specific command). 
 
 The function and telecommand was initially designed to be scale-able for multiple commands, hence why there was initially a telecommand to only disable one GNSS subsystem, but due to the small list of things we can actually enable/disable, it was axed. We can discuss further tomorrow in person, responding here so I don't forget:) Thanks for the feedback!

NuclearTea:
@celinahy Your PR is failing the github action, please take a look

DeflateAwning:
This PR is made obsolete by rewriting the DMC3 to support EPS-based power control (code updates in #376). 

Thanks for everyone who worked on this! I'm glad we were able to arrive at a more elegent and simpler solution than these software-based control mechanisms.

As far as I can tell, none of the changes in this PR are required, and this PR can now be closed without merging.

Issue #117, opened 2024-07-20
Log a warning if the system time changes by too much


JohnathanBurchill:
Warn if time synchronization changes the system time by more than some threshold, like 2 s, say.

DeflateAwning:
Also, on each time sync, log (info-level) the old time (sync_source_offset format), new time (sync_source_offset, where offset is approx. 0), as well as the signed clock shift as a result.

DeflateAwning:
The solution is <10-20 lines, and should go in `TIM_set_current_unix_epoch_time_ms`, I think.

DeflateAwning:
Implemented in #323

PR #323, opened 2025-04-07
Logs for clock time syncs

Ozone-Whiskey:
Addresses #117. Implemented LOG warn message for time syncs over 2000ms and LOG info about time before sync, time after sync, and time correction in offset format. Needs testing.

### DIFF
--- a/firmware/Core/Src/timekeeping/timekeeping.c
+++ b/firmware/Core/Src/timekeeping/timekeeping.c
@@ -63,7 +63,7 @@ void TIM_set_current_unix_epoch_time_ms(uint64_t current_unix_epoch_time_ms, TIM
     if(correction_time_ms>=2000 || correction_time_ms<=-2000)
     {
         LOG_message(
-            LOG_SYSTEM_ALL, LOG_SEVERITY_WARNING, LOG_SINK_ALL,
+            LOG_SYSTEM_OBC, LOG_SEVERITY_WARNING, LOG_SINK_ALL,
             "Synchronization has changed system time by 2000ms or more. Time deviation was %s ms.",
             correction_time_ms_str
         );
@@ -74,7 +74,7 @@ void TIM_set_current_unix_epoch_time_ms(uint64_t current_unix_epoch_time_ms, TIM
 
     // Log both times and the correction
     LOG_message(
-        LOG_SYSTEM_ALL, LOG_SEVERITY_DEBUG, LOG_SINK_ALL,
+        LOG_SYSTEM_OBC, LOG_SEVERITY_NORMAL, LOG_SINK_ALL,
         "Time synchronized. Old time: %s, New time: %s, Clock shift: %s ms.",
         old_time_str, new_time_str, correction_time_ms_str
     );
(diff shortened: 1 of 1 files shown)


firmware/Core/Src/timekeeping/timekeeping.c
```diff
@@ -27,6 +27,14 @@ void TIM_set_current_unix_epoch_time_ms(uint64_t current_unix_epoch_time_ms, TIM
     // Don't log it right away; update the time info, then log after.
     const uint8_t is_this_sync_before_the_last_sync = current_unix_epoch_time_ms < TIM_unix_epoch_time_at_last_time_resync_ms;
 
+    // Compute time before applying sync
+    uint64_t unix_epoch_time_before_sync_ms = TIM_get_current_unix_epoch_time_ms();
```
DeflateAwning, line 31: Please mark variables which do not change value as `const`.

DeflateAwning (changes requested):
Looks great, thanks! Just a tiny tweak.

Can you explain how you've tested this, perhaps with an example?

Ozone-Whiskey:
> Looks great, thanks! Just a tiny tweak.
> 
> Can you explain how you've tested this, perhaps with an example?

I tested the 2000ms warning with a telecommand check and it worked perfectly (ended up removing it from timekeeping telecommands file).

The time sync log logic was tested on a separate C file with just code relevant to the log like the timestamp function. I used https://www.unixtimestamp.com/ for example timestamps with just an example source.

The time sync hardware testing will be done tomorrow in the clean room with hardware, so it is untested on hardware.

DeflateAwning:
Just do confirm, do you mean like this:

1. Send tcmd `CTS1+set_system_time(1234123123)!`
2. Send tcmd `CTS1+set_system_time(1234123124)!` again, less than 2 seconds later.
3. Observe that no warning happens (just the normal sync log message).
4. Wait 10 seconds, and then send that same command again: `CTS1+set_system_time(1234123124)!`
5. Observe the warning.
6. Wait 10 seconds, then try setting the time to before the last sync time: `CTS1+set_system_time(1234123120)!`
7. Observe the other log message which mentions the problem with a sync time less than the previous sync time (logs are unsortable). 

This is the type of test procedue that can be defined quickly before/during testing so it's clear what cases have and haven't been addressed :)

DeflateAwning (dismissed):

Ozone-Whiskey:
> Just do confirm, do you mean like this:
> 
> 1. Send tcmd `CTS1+set_system_time(1234123123)!`
> 2. Send tcmd `CTS1+set_system_time(1234123124)!` again, less than 2 seconds later.
> 3. Observe that no warning happens (just the normal sync log message).
> 4. Wait 10 seconds, and then send that same command again: `CTS1+set_system_time(1234123124)!`
> 5. Observe the warning.
> 6. Wait 10 seconds, then try setting the time to before the last sync time: `CTS1+set_system_time(1234123120)!`
> 7. Observe the other log message which mentions the problem with a sync time less than the previous sync time (logs are unsortable).
> 
> This is the type of test procedue that can be defined quickly before/during testing so it's clear what cases have and haven't been addressed :)

Tested on hardware using this procedure and it worked! New commit just updated log source and severity.

NuclearTea (approved):
Nice! This was tested and passed! The only changes after were changing the Log source and severity

DeflateAwning (approved):
Thanks! Please merge!

Issue #324, opened 2025-04-07
Bug: LOG_message early exit if debugging is disabled doesn't work


DeflateAwning:
There's an out-of-bounds read in this part of the code, I think:

```c
void LOG_message(LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_t sink_mask, const char fmt[], ...)
{
    // Ensure quick return if debugging is disabled
    // Needed to maintain good hot-path performance
    if (severity == LOG_SEVERITY_DEBUG) {
        // Return early if debugging is not enabled for this system
        if (!(severity & LOG_systems[1 << source].severity_mask)) { // <- Index into LOG_systems is WAY out of bounds
            return;
        }
        // Return early if debugging is not enabled for ALL of the requested sinks
        uint8_t debugging_enabled_for_at_least_one_sink = 0;
        for (uint16_t i = 0; i < LOG_NUMBER_OF_SINKS; i++) {
            if (LOG_sinks[i].enabled) {
                debugging_enabled_for_at_least_one_sink = 1;
                break;
            }
        }
        if (!debugging_enabled_for_at_least_one_sink) {
            return; 
        }
    }
    // ...
}

PR #333, opened 2025-04-14
fix: LOG_message fails to show debug messages in certain cases

destinsaba:
Fixes issue #324 by correcting the way we indexed into the LOG_systems array. Commit history contains a test_log.c file with some simple tests showing the indexing before vs. after.


DeflateAwning (approved):
This looks awesome! Very simple solution, nice.

Stumbled across a source that says `__builtin_ctz(0)` (i.e., calling with argument 0) is undefined behaviour. I think there's currently no case where that happens though, so should be fine.

Just a couple notes on process:
1. Please open branch names with the format `your-name/less-than-5-words-describing-the-purpose-i324`
2. Please make the title of the pull request be the clean description of the change (because it becomes the commit message in the `main` branch).

Congrats on your first contribution. Feel free to add your name to the contributor list in the README (either in this PR or your next)! 

Click "Squash and Merge" when you're ready to merge.

PR #334, opened 2025-04-18
Reading MPI Science Mode Enable / Disable

Saksham-P:
This allows us to prepare the system to be ready for reading and storing MPI science data.

Preparation function:
- Mount LFS if not already mounted
- Create a new file with current HAL_Tick appended on the file name to be used as storage
- Open the file and set the readiness state to 1

Enable active function:
- Set system state to accept data from MPI
- Open DMA channel
- If DMA channel opened successfully, set the readiness state to 2 (indicating we can now close the file when active mode is disabled)

Thread functionailty:
- Write to the open file if there is data to be written 
- Close the file if system not in sensing mode anymore and readiness state is 2
- **Currently has debug / log statement, which will be removed after a final review**

### DIFF
--- a/firmware/Core/Src/mpi/mpi_command_handling.c
+++ b/firmware/Core/Src/mpi/mpi_command_handling.c
@@ -266,9 +266,6 @@ uint8_t MPI_enable_active_mode(const char MPI_science_file_name[]) {
 
     MPI_set_transceiver_state(MPI_TRANSCEIVER_MODE_MISO); // Set the MPI transceiver to MISO mode
     MPI_current_uart_rx_mode = MPI_RX_MODE_SENSING_MODE;
-
-    // FIXME: Do we need this? Had this before to fix an issue now fixed by AbortReceive
-    __HAL_UART_CLEAR_OREFLAG(UART_mpi_port_handle);
     
     // Receive MPI response actively with 8192 buffer size.
     const HAL_StatusTypeDef rx_status = HAL_UART_Receive_DMA(
--- a/firmware/Core/Src/mpi/mpi_command_handling.c
+++ b/firmware/Core/Src/mpi/mpi_command_handling.c
@@ -1,9 +1,11 @@
 #include "mpi/mpi_command_handling.h"
 #include "mpi/mpi_types.h"
 #include "mpi/mpi_transceiver.h"
+#include "eps_drivers/eps_channel_control.h"
 #include "uart_handler/uart_handler.h"
 #include "log/log.h"
 #include "debug_tools/debug_uart.h"
+#include "littlefs/littlefs_helper.h"
 
 #include "main.h"
 #include "stm32l4xx_hal.h"
@@ -19,7 +21,22 @@ static const uint16_t MPI_RX_TIMEOUT_DURATION_MS = 2000;
 
 volatile MPI_rx_mode_t MPI_current_uart_rx_mode = MPI_RX_MODE_NOT_LISTENING_TO_MPI;
 
+/// @brief Current state of the `MPI_active_data_median_buffer` (pending write vs. written).
+volatile MPI_buffer_state_enum_t MPI_buffer_state = MPI_MEMORY_WRITE_STATUS_READY;
 
+uint8_t MPI_receive_prepared = 0;
+uint8_t MPI_science_data_file_is_open = 0;
+lfs_file_t MPI_science_data_file_pointer;
+
+/// @brief Sends commandcode+params to the MPI as bytes
+/// @param bytes_to_send Buffer containing the telecommand + params (IF ANY) as hex bytes
+/// @param bytes_to_send_len Size of telecommand buffer
+/// @param rx_buffer Buffer to store incoming response from the MPI
+/// @param rx_buffer_max_size The maximum size of the MPI response buffer
+/// @param rx_buffer_len Pointer to variable that will contain the length of the populated MPI response buffer 
+/// @return 0: Success, 2: Failed UART transmission, 3: Failed UART reception, 
+///         4: Timeout waiting for 1st byte from MPI, 8: Not enough space in the MPI response buffer
+/// @note If the MPI is in "science data" mode, it will be disabled after the command is executed.
 uint8_t MPI_send_command_get_response(
     const uint8_t *bytes_to_send, const size_t bytes_to_send_len, 
     uint8_t *rx_buffer, const size_t rx_buffer_max_size, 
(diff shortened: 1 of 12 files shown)


NuclearTea (commented):
Great Start!

Just some small things that need to be touched up.

We need to plan a time with Dr.Burchill on testing this lol.

firmware/Core/Src/mpi/mpi_command_handling.c
```diff
@@ -144,3 +152,82 @@ uint8_t MPI_validate_command_response(
...
+    else if (receive_status == HAL_TIMEOUT) {
+        return 3; // Error code: Timeout while trying to set DMA
+    }
+    else if (receive_status == HAL_ERROR) {
+        MPI_set_transceiver_state(MPI_TRANSCEIVER_MODE_INACTIVE);
+        HAL_UART_DMAStop(UART_mpi_port_handle);
+        MPI_current_uart_rx_mode = MPI_RX_MODE_NOT_LISTENING_TO_MPI;
+        return 2; // Error code: Failed UART reception
+    }
+
```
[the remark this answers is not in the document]
Saksham-P: I've pushed some changes, but I've kept it so that if the UART line is busy when trying to enable the DMA on the same UART line, we don't change anything because I'm worried it might interfere with another process that might be using the UART line.

Saksham-P:
An important thing to note for the MPI thread being used is that there is only 1 second of delay before the main while loop begins. 

```
void TASK_service_write_mpi_data(void *argument) {
    TASK_HELP_start_of_task();
    osDelay(1000); <--------------
    
    uint32_t start_time;

    while(1) {
        ...
```

Not sure how long we want this delay to be exactly, but I am assuming > 15 seconds?

Saksham-P:
With the latest changes made to the code, I found some interesting bugs which I am not sure how to solve. 

They are mainly issues with how DMA and UART works with receiving data.

**MPI sends data when DMA port isn't open / listening for data**
If the `UART_mpi_port_handle` isn't set to receive data yet and the mpi is still sending data over to the UART lines. This causes an issue where once the `UART_mpi_port_handle` is set to receive data using `HAL_UART_Receive_DMA`, the data doesn't come through and neither the half or complete callback are called. 

**UART / DMA port stops receiving while MPI is actively sending data**
If the MPI is actively sending data to the UART lines, and we use `HAL_UART_DMAStop`. This causes a similar issue as explained above where once we set the UART line to receive data again using `HAL_UART_Receive_DMA`, the data doesn't come through and neither the half or the complete callbacks are triggered. 

I am not sure why the above issues are happening, and I don't currently have a solution. If @NuclearTea @DeflateAwning you guys have seen something similar happen, I would like to know. 

But I have somewhat of an idea of a workaround. We could possibly always keep the DMA to receive data from the mpi uart line, but only look at the data once we need to (when mpi science mode gets enabled). I haven't tested this idea out, but in theory this would allows us to not have to deal with the issues I mentioned above.

JohnathanBurchill:
Cool - we will test this out with the MPI this coming week.

NuclearTea (commented):
Just a small comment, looks promising otherwise

DeflateAwning:
FYI - Rebased and pushed to `da/mpi-science-mode` branch

DeflateAwning:
Testing mostly good in that other branch. Nice.

DeflateAwning:
Pushed a backup at this point - `da/backup-mpi-science-mode-20250507`

DeflateAwning:
Squashed, rebased, and force-pushed back to this branch/PR. `da/mpi-science-mode` is now obselete and everything is on here. 

@Saksham-P, can you please pull this version, and modify it so that the "start" command takes a filename argument? Then, I think this is mostly good! Going to do a quick review with comments right now, but sounds like testing is going quite well. Thanks a lot :)

firmware/Core/Src/mpi/mpi_command_handling.c
```diff
@@ -128,3 +145,168 @@ uint8_t MPI_validate_command_response(
...
+    if (open_result < 0) {
+        LOG_message(LOG_SYSTEM_MPI, LOG_SEVERITY_WARNING, LOG_all_sinks_except(LOG_SINK_FILE), "Error opening / creating file: %s", MPI_science_data_file_name);
+        MPI_receive_prepared = 0;
+        return open_result;
+    }
+    MPI_science_data_file_is_open = 1;
+    LOG_message(LOG_SYSTEM_MPI, LOG_SEVERITY_NORMAL, LOG_all_sinks_except(LOG_SINK_FILE), "Opened/created file: %s", MPI_science_data_file_name);
+    MPI_receive_prepared = 1;
+    
+    // Total 5 second delay to make sure MPI is booted
```
JohnathanBurchill, line 224: What happens if the delay is negative? Suggest to remove this possibility by assigning the calculated value to a signed integral type, then delaying only if positive.

Saksham-P: Good catch! I'll get that updated tonight.

There are also a few things I want to clarify I will add onto my next comment after updating the code.

JohnathanBurchill:
We tested an earlier version of this with the spare OBC and MPI last week, confirming storing of telemetry, and saw of order 1 MByte stored to the filesystem. We did not check the contents of the file. We should test this latest version with the MPI before merging into main.

DeflateAwning:
Just rebased and force-pushed. Also tacking on updates to the review comments so it's ready to test tomorrow!

firmware/Core/Src/mpi/mpi_command_handling.c
DeflateAwning: Appears implemented. Good catch!

firmware/Core/Src/uart_handler/uart_handler.c
```diff
@@ -331,7 +331,7 @@ void HAL_UART_RxHalfCpltCallback(UART_HandleTypeDef *huart) {
                 // DEBUG_uart_print_str("HALF - Received 4096 Bytes!\n");
             } else {
                 DEBUG_uart_print_str("HALF - *Assumed* 4096 Bytes are being lost!\n");
```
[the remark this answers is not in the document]
DeflateAwning: I fixed just now.

firmware/Core/Src/uart_handler/uart_handler.c
```diff
@@ -312,8 +330,30 @@ void HAL_UART_RxHalfCpltCallback(UART_HandleTypeDef *huart) {
         CAMERA_uart_half_1_state = CAMERA_UART_WRITE_STATE_HALF_FILLED_WAITING_FS_WRITE;
         CAMERA_uart_half_2_state = CAMERA_UART_WRITE_STATE_HALF_FILLING;
     }
-}
 
+    else if (huart->Instance == UART_mpi_port_handle->Instance) {
```
NuclearTea, line 334: I believe you already implement this case above. Please remove this to remove confusion

DeflateAwning: I don't follow; there's code in both the full callback and half-callback for the UART_mpi_port_handle. Is that what you're seeing?

NuclearTea: Apologies, github's truncated view was confusing me. There is no issue here!

DeflateAwning:
I'm going to close this in favor of the new PR #463

PR #298, opened 2025-03-20
Create telecommand: `system_self_check_failures_as_json`

DeflateAwning:

PR #337, opened 2025-04-18
OBC Temperature Sensor: Simplify usage, review compliance

DeflateAwning:
This PR pulls out the GNSS-related changes from PR https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/pull/298 (self checks), as that PR was getting a bit all-over-the-place.

Main feature-enrichment contribution: Simple-to-use `OBC_TEMP_SENSOR_get_temperature_cC` function, which is the simplest function to use for reading the OBC temperature.


DeflateAwning:
Looks good to me! Verbose test record below. LFT temperature sensor reports 28C, so 2900 cC makes sense.

[obc_temp_test_record.txt](https://github.com/user-attachments/files/19861612/obc_temp_test_record.txt)

PR #340, opened 2025-04-23
ADCS: assorted bugfixes and improvements from testing

DeemDeem52:


firmware/Core/Inc/adcs_drivers/adcs_internal_drivers.h
```diff
@@ -7,6 +7,7 @@
 #include <time.h>
 
 #define ADCS_i2c_HANDLE &hi2c1
+#define WATCHDOG &hiwdg
```
DeflateAwning, line 10: Feel unnecessary imo; there's only one watchdog

DeflateAwning:
I'm willing to be a bit flexible with it, but I'd prefer to document what the bugs are as GitHub Issues, with at least a tiny description. Then, we can have a PR that closes that issue/bug (or set of bugs, if somewhat related). Nbd if you really don't want to do that, but just wanted to throw it out as I've requested it of others as well.

Documenting what bugs made it through QA into the main branch is helpful for several reasons, including minimizing minimizing forgotten bugs, auditing failures and bugs, informing future satellite missions, and as training material.

DeflateAwning (approved):
Everything seems reasonable! Approving on the basis that these changes have been tested, and are otherwise correct.

Thanks!

DeemDeem52:
Thanks Parker! Glad I could help out!

PR #349, opened 2025-04-25
fix: Prevent LFS formatting if LFS is mounted

Saksham-P:
I am hoping to find any potential bugs or edge cases for LFS code and then adding fixes here. 

Currently I only have 1 which stops users from formatting FLASH memory if LFS is mounted on the memory. 
This has the potential of corrupting the FLASH memory or making any proceeding operations fail.

- We can keep it like this stopping the format 
- Or, we can unmount the LFS in code and continue with format

I will continue adding fixes as soon as I find them

### DIFF
--- a/firmware/Core/Src/littlefs/littlefs_helper.c
+++ b/firmware/Core/Src/littlefs/littlefs_helper.c
@@ -58,6 +58,12 @@ struct lfs_file_config LFS_file_cfg = {
 /// @return 0 on success, negative LFS error codes on failure
 int8_t LFS_format()
 {
+    if (LFS_is_lfs_mounted) {
+        LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_WARNING, LOG_all_sinks_except(LOG_SINK_FILE), 
+        "FLASH Memory cannot be formatted while LFS is mounted!");
+        return 1;
+    }
+    
     int8_t format_result = lfs_format(&LFS_filesystem, &LFS_cfg);
     if (format_result < 0) {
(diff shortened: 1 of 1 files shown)


DeflateAwning:
Awesome! If you're good with it, I'd be happy to approve it right away so it's usable for everyone else right away. 

Having separate issues (if applicable), branches, and PRs for each bug, so that each commit describes the bug/improvement, is ideal. I know it's a bit of extra branch-switching, but is optimal

Saksham-P:
Sounds good!

DeflateAwning:
Docstring says:
```
/// @return 0 on success, negative LFS error codes on failure
```

Should add that it also returns positive error codes too

Saksham-P:
I also noticed for the function `lfs_ssize_t LFS_file_size(const char file_name[])` we are doing the LFS mounted check and returning -12512 if LFS not mounted. Is that a specific value or just a random negative value?
```
lfs_ssize_t LFS_file_size(const char file_name[])
{
    if (!LFS_is_lfs_mounted) {
        LOG_message(LOG_SYSTEM_LFS, LOG_SEVERITY_CRITICAL, LOG_all_sinks_except(LOG_SINK_FILE), "LittleFS not mounted");
        return -12512;
    } 
```

Line 537 in `littlefs_helper.c`

Saksham-P:
Sorry, disregard the previous comment, but I do want to know if we should be returning the number of bytes written for LFS_write functions, currently it's 0 for no errors and <0 for errors, 1 if unmounted

DeflateAwning:
It's a random negative number. Per the docstring, `LFS_file_size()` returns negative numbers on failure and positive numbers for filesizes.

DeflateAwning (dismissed):
Thanks!

DeflateAwning (approved):

PR #359, opened 2025-04-27
Toggle which sources can Log message

2BlackCats:
Sorry I can't test this. If someone can test it on their STM.

### DIFF
--- a/firmware/Core/Inc/log/log.h
+++ b/firmware/Core/Inc/log/log.h
@@ -52,8 +52,6 @@ enum {
     LOG_SYSTEM_ON = 1,
 };
 
-extern uint16_t LOG_sources_enabled;
-
 void LOG_message(
     LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_t sink_mask,
     const char fmt[], ...
--- a/firmware/Core/Inc/log/log.h
+++ b/firmware/Core/Inc/log/log.h
@@ -52,7 +52,7 @@ enum {
     LOG_SYSTEM_ON = 1,
 };
 
-extern uint8_t LOG_toggled_sources[15];
+extern uint16_t LOG_sources_enabled;
 
 void LOG_message(
     LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_t sink_mask,
--- a/firmware/Core/Inc/log/log.h
+++ b/firmware/Core/Inc/log/log.h
@@ -52,6 +52,8 @@ enum {
     LOG_SYSTEM_ON = 1,
 };
 
+extern uint8_t LOG_toggled_sources[15];
+
 void LOG_message(
     LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_t sink_mask,
     const char fmt[], ...
@@ -76,5 +78,6 @@ uint8_t LOG_memory_table_max_entries(void);
 uint8_t LOG_get_memory_table_index_of_most_recent_log_entry(void);
 const char *LOG_get_memory_table_full_message_at_index(uint8_t index);
 const char *LOG_get_most_recent_log_message_text(void);
+LOG_system_enum_t LOG_source_from_str(const char source_name[]);
 
 #endif // __INCLUDE__GUARD__LOG_H_
--- a/firmware/Core/Src/log/log.c
+++ b/firmware/Core/Src/log/log.c
@@ -12,24 +12,7 @@
 
 // Inspired by uLog: https://github.com/rdpoor/ulog
 
-uint8_t LOG_toggled_sources[15] = 
-{
-    1, // LOG_SYSTEM_OBC
-    1, // LOG_SYSTEM_UHF_RADIO
-    1, // LOG_SYSTEM_UMBILICAL_UART
-    1, // LOG_SYSTEM_GPS
-    1, // LOG_SYSTEM_MPI
-    1, // LOG_SYSTEM_EPS
-    1, // LOG_SYSTEM_BOOM
-    1, // LOG_SYSTEM_ADCS
-    1, // LOG_SYSTEM_LFS
-    1, // LOG_SYSTEM_FLASH
-    1, // LOG_SYSTEM_ANTENNA_DEPLOY
-    1, // LOG_SYSTEM_LOG
-    1, // LOG_SYSTEM_TELECOMMAND
-    1, // LOG_SYSTEM_UNIT_TEST
-    1, // LOG_SYSTEM_ALL
-};
+uint16_t LOG_sources_enabled  = 0xFFFF; // All sources are enabled by default
 
 // Internal interfaces and variables
 #define LOG_TIMESTAMP_MAX_LENGTH 30
@@ -128,7 +111,7 @@ static const uint16_t LOG_NUMBER_OF_SYSTEMS = sizeof(LOG_systems) / sizeof(LOG_s
 ///     Exclude one or more sinks using LOG_all_sinks_except(...)
 void LOG_message(LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_t sink_mask, const char fmt[], ...)
 {
-    if (LOG_toggled_sources[source] == 0) {
+    if (!(source & LOG_sources_enabled)) {
         return; // Logging is disabled for this source
     }
     // Ensure quick return if debugging is disabled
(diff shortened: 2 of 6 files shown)


firmware/Core/Src/log/log.c
```diff
@@ -12,6 +12,25 @@
 
 // Inspired by uLog: https://github.com/rdpoor/ulog
 
+uint8_t LOG_toggled_sources[15] =
```
DeflateAwning, line 15: Should this be named `LOG_sources_enabled` ? Assuming it contains a list of bool-like uint8_t of which ones are enabled?

firmware/Core/Src/log/log.c
```diff
@@ -453,3 +475,42 @@ const char *LOG_get_most_recent_log_message_text(void)
...
+    if (strcmp(source_name, "9") == 0) return LOG_SYSTEM_FLASH;
+    if (strcmp(source_name, "10") == 0) return LOG_SYSTEM_ANTENNA_DEPLOY;
+    if (strcmp(source_name, "11") == 0) return LOG_SYSTEM_LOG;
+    if (strcmp(source_name, "12") == 0) return LOG_SYSTEM_TELECOMMAND;
+    if (strcmp(source_name, "13") == 0) return LOG_SYSTEM_UNIT_TEST;
+    if (strcmp(source_name, "14") == 0) return LOG_SYSTEM_ALL;
+
+    
+    // Parse Strings.
+    if (strcmp(source_name, "OBC") == 0) return LOG_SYSTEM_OBC;
```
DeflateAwning, line 499: Please use `strcasecmp` instead of `strcmp` in here to do a case-insensitive compare. Your docstring in the other function is currently invalid (it says the lowercase values). Case-insensitive is best though.

firmware/Core/Src/log/log.c
```diff
@@ -453,3 +475,42 @@ const char *LOG_get_most_recent_log_message_text(void)
...
+    // Parse Numbers.
+    if (strcmp(source_name, "0") == 0) return LOG_SYSTEM_OBC;
+    if (strcmp(source_name, "1") == 0) return LOG_SYSTEM_UHF_RADIO;
+    if (strcmp(source_name, "2") == 0) return LOG_SYSTEM_UMBILICAL_UART;
+    if (strcmp(source_name, "3") == 0) return LOG_SYSTEM_GPS;
+    if (strcmp(source_name, "4") == 0) return LOG_SYSTEM_MPI;
+    if (strcmp(source_name, "5") == 0) return LOG_SYSTEM_EPS;
+    if (strcmp(source_name, "6") == 0) return LOG_SYSTEM_BOOM;
+    if (strcmp(source_name, "7") == 0) return LOG_SYSTEM_ADCS;
+    if (strcmp(source_name, "8") == 0) return LOG_SYSTEM_LFS; // Engg Model ADCS
```
DeflateAwning, line 489: Please remove the comment at the end of this line haha

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -274,3 +274,46 @@ uint8_t TCMDEXEC_log_report_n_latest_messages_from_memory(const char *args_str,
...
 }
+
+/// @brief Telecommand: Toggle the logging state of a source
+/// @param args_str
+/// - Arg 0: Source enum
+/// - Arg 1: State 0: disable source logging, 1: enable source logging
+/// @note Valid string values for Arg 0: "obc", "uhf_radio", "umbilical_uart", "gps",
+///     "mpi", "eps", "boom", "adcs", "lfs", "flash", "antenna_deploy", "log",
+///     "telecommand", "unit_test"
+uint8_t TCMDEXEC_toggle_source_logging(const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
```
DeflateAwning, line 285: Please rename to `TCMDEXEC_log_set_logging_source_enabled`

firmware/Core/Inc/log/log.h
```diff
@@ -52,6 +52,8 @@ enum {
     LOG_SYSTEM_ON = 1,
 };
 
+extern uint8_t LOG_toggled_sources[15];
```
DeflateAwning, line 55: Can we store this as a single uint32_t bitfield to be consistent with the code that's already written please?

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -274,3 +274,46 @@ uint8_t TCMDEXEC_log_report_n_latest_messages_from_memory(const char *args_str,
...
+
+    LOG_system_enum_t LOG_source = LOG_source_from_str(source);
+    if (LOG_source == LOG_SYSTEM_UNKNOWN) {
+        snprintf(
+            response_output_buf, response_output_buf_len,
+            "Error parsing channel arg: Unknown source %s", source);
+        return 2;
+    }
+    
+    for (uint16_t i = 0; LOG_source >> i; i++) {
```
DeflateAwning, line 312: When you store the `LOG_toggled_sources` variable as a single uint32_t bitfield, you'll see that this block becomes much simpler and doesn't have a loop.

firmware/Core/Src/log/log.c
```diff
@@ -109,6 +128,9 @@ static const uint16_t LOG_NUMBER_OF_SYSTEMS = sizeof(LOG_systems) / sizeof(LOG_s
 ///     Exclude one or more sinks using LOG_all_sinks_except(...)
 void LOG_message(LOG_system_enum_t source, LOG_severity_enum_t severity, uint32_t sink_mask, const char fmt[], ...)
 {
+    if (LOG_toggled_sources[source] == 0) {
```
DeflateAwning, line 131: Pretty sure this doesn't work, and results in a massive out-of-bounds memory read. The value which `source` contains is like `1 << 12` (4096) - which is a bad index to use to access a 15-element array. 

A simple test would have found this issue.

DeflateAwning (changes requested):
Overall comments:

* Can you please ensure we call them "log systems" instead of "log sources". While I agree that "log source" might make a bit more sense, remaining consistent with the majority of the code that exists makes code easier to work with.

* Please use a bitfield to store the enable states instead of an array.

2BlackCats:
Log Systems seems to be directed at Logging to LFS so I think it still makes sense to keep them Log Source. Also made it a uint16_t bit field since there were only 15 different systems. If you still want it changed I can. Just a reminder to test on a STM before you approve. THX

DeflateAwning:
I think the prefered way to do this is to make the telecommand call `LOG_set_system_severity_mask`, instead of introducing new variables. 

Would you mind updating it to do it like that? Should result in a smaller change that's compatible with the "report logging states" functions that exist. Thanks!

2BlackCats:
I changed the use of the previously created telecommand to set the severity mask of the system. Using  LOG message now checks if the source LOG system is on and its correlated severity of logging is enabled.

firmware/Core/Src/log/log.c
```diff
@@ -109,6 +109,22 @@ static const uint16_t LOG_NUMBER_OF_SYSTEMS = sizeof(LOG_systems) / sizeof(LOG_s
...
+            break;
+        }
+    }
+    if (LOG_source == NULL) {
+        // Source not found
+        LOG_message(LOG_SYSTEM_LOG, LOG_SEVERITY_ERROR, LOG_SINK_ALL, "LOG_message(): unknown source: %d", source);
+        return;
+    }
+
+    if (LOG_source->file_logging_enabled && (severity & LOG_source->severity_mask)) {
```
DeflateAwning, line 125: This seems like an unnecessary loop. Can you do it alongside the `LOG_system_t *system = &LOG_systems[LOG_NUMBER_OF_SYSTEMS - 1];` part below? 

Maybe the `vsnprintf` part could be moved to be below your "is severity enabled for this log source" check to increase performance? 

This is looking better!

firmware/Core/Src/telecommands/log_telecommand_defs.c
```diff
@@ -274,3 +274,57 @@ uint8_t TCMDEXEC_log_report_n_latest_messages_from_memory(const char *args_str,
...
+    }
+
+    if (severity_mask > 32) {
+        snprintf(
+            response_output_buf, response_output_buf_len,
+            "Error parsing severity_mask arg: Severity mask %d is too large", (uint16_t) severity_mask);
+        return 2;
+    }
+
+    LOG_system_enum_t LOG_source = LOG_source_from_str(source);
```
DeflateAwning, line 317: Please mark all unmutabled variables as `const` in this block.

JohnathanBurchill:
Tried testing this on a dev kit. I'm getting no TM through the UART. Might help to rebase now to main, which does work on the dev kit and which generates lots of EPS error messages to test with. :)

DeflateAwning:
Superseded by #481

Thank you for working on this

Issue #365, opened 2025-04-29
Update MPI firmware to match ICD Rev 1.4 version


DeflateAwning:
Updates to ICD:
* On bootup, the MPI does not start sending science data. You must send the "start science data" command (`{0x54, 0x43, 0x13}`) to begin.
* To stop science data, send (`0x54, 0x43, 0x14`)
* On command successful, the MPI sends back [`<command code char after TC>`, `0xFE`] (where 0xFE = 254). On error, there is any other value. Response value 0x00 is not sent anymore on success.

This update was codified on 2025-05-01 in ICD Rev 1.4.

DeflateAwning:
This is implemented in PR #366 and PR #334

PR #366, opened 2025-04-29
Fix MPI command mode, update to MPI ICD Rev 1.4

DeflateAwning:
Related to #365


DeflateAwning:
Stuff mostly good here. Will probably just remove the single-byte command mode DMA in favor of the new simple blocking mode one.

TODO: Check timeouts on the blocking mode one. Seemed to quit out quickly.

TODO: Maybe split out the duplex mode.

TODO (New issue probably): Add a "send command to MPI" without reception (and for all UART PORTS)

DeflateAwning:
Got a report that testing this went well today. Purged the parts that aren't working well. Need to check over the self-test behaviour a bit still, but seems ready enough to merge. 

TODO: Still need to check the self-test functionality.

PR #368, opened 2025-04-30
Telecommand to get state of RBF pin

2BlackCats:


2BlackCats:
I dont have a kit and can't test. If someone could test it as well.

firmware/Core/Src/telecommands/system_telecommand_defs.c
```diff
@@ -153,3 +153,14 @@ uint8_t TCMDEXEC_system_self_check_failures_as_json(
     );
     return 0;
 }
+
+uint8_t TCMDEXEC_obc_get_rbf(
+    const char *args_str, TCMD_TelecommandChannel_enum_t tcmd_channel,
+    char *response_output_buf, uint16_t response_output_buf_len
+) {
+    GPIO_PinState rbf_state = HAL_GPIO_ReadPin(PIN_REMOVE_BEFORE_FLIGHT_LOW_IS_FLYING_IN_GPIO_Port,
```
DeflateAwning, line 161: Please mark as const.

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -123,6 +123,12 @@ const TCMD_TelecommandDefinition_t TCMD_telecommand_definitions[] = {
         .number_of_args = 0,
         .readiness_level = TCMD_READINESS_LEVEL_FOR_OPERATION,
     },
+    {
+        .tcmd_name = "obc_get_rbf",
```
DeflateAwning, line 127: Please rename to `obc_get_rbf_state`

DeflateAwning (changes requested):
Please make a function called `OBC_read_rbf_state` which returns an enum.

The enum values should be `OBC_RBF_STATE_BENCH` and `OBC_RBF_STATE_FLYING` (mentioned in the issue).

Then, the response should say "BENCH" or "FLYING" in the JSON (perhaps in addition to the binary state as currently implemented).

Thanks!

DeflateAwning (approved):
Thanks

PR #371, opened 2025-05-01
Asjad cam bps - DRAFT

MrBear101:
in its current state camera_setup() needs to be run before every image as it handles a lot of the setup steps and sets the filename. Capturing image has been iffy so far but has worked to an extent. Baudrate 460800 CANNOT be used


NuclearTea:
Closing pull request as the branch that will be merged is `asjad-cam-bps-with-watchdog`

PR #379, opened 2025-05-07
ADCS: Free up some unused telecommands

DeemDeem52:


DeemDeem52:
Since the ADCS doesn't have a star tracker and we're not using the GPS functions of the ADCS, we can free up 6 telecommand slots without really any effort at all.
(I might take one slot back later to write a function to erase files from the ADCS SD card, though, since that came up during testing.)

PR #383, opened 2025-05-08
Purge libcsp: Delete all `Core/Src/csp/` and `Core/Inc/csp/`

DeflateAwning:
libcsp is no longer needed, as the functionality vs. size tradeoff did not make sense, and the features from it (e.g., generating the CSP header) are trivial to implement otherwise


parker-research:
Clean build main at commit 20350008 (includes libcsp)

```
Memory region         Used Size  Region Size  %age Used
             RAM:      483104 B       640 KB     73.72%
            RAM2:          0 GB        64 KB      0.00%
            RAM3:          0 GB       384 KB      0.00%
           FLASH:      329080 B         2 MB     15.69%
   text    data     bss     dec     hex filename
 327312    1768  481352  810432   c5dc0 build/debug/CTS-SAT-1_FW.elf
```

Clean build main after removing libcsp:

```
Memory region         Used Size  Region Size  %age Used
             RAM:      483104 B       640 KB     73.72%
            RAM2:          0 GB        64 KB      0.00%
            RAM3:          0 GB       384 KB      0.00%
           FLASH:      329072 B         2 MB     15.69%
   text    data     bss     dec     hex filename
 327304    1768  481352  810424   c5db8 build/debug/CTS-SAT-1_FW.elf
```

Seems that removing it does not decrease the firmware build size significantly. Still worth removing though.

parker-research (approved):

Issue #8, opened 2024-03-10
Telecommand: take a photo


DeflateAwning:
Command should take a photo and store it to the FS.

Args:
1. Light setting

* Please review the camera datasheet spec to confirm if there are any other settings which can be args.

MrBear101:
Additional Arg:

    2. Flash setting
 
Current Progress:

    Camera was successfully interfaced with STM32 board.
    Command was sent and Image was received in it's entirety several times through UART.
    Currently merging camera code with a clone of the Firmware repo.
    
Next Steps:

    Will begin interfacing camera with memory module and testing storage of photos using littleFS

DeflateAwning:
Can you please get your code in a branch in this repo, and ensure you've pushed that branch?

DeflateAwning:
@MrBear101, you should be able to start working on this pretty easily! The filesystem code is in the `main` branch now, so it's just a matter of calling into those functions! Let me or Sakshum know if you have any issues with the filesystem! 

If you need any IOC configuration changes (e.g., setting interrupts or DMA), let me know! Make sure not to commit those changes to your branch.

Hoping to see some progress on this project this week!

DeflateAwning:
@MrBear101, can you please provide an update on the progress of this task?

Ozone-Whiskey:
@MrBear101 How is thing going?

robotoshi:
Ali and Asjad tested and I talked with them today.

@MrBear101 @NuclearTea

We need to add a new RTOS task that handles camera DMA data (and possibly other data in large volumes).
Priority should be higher than "normal".
Should run a `while(1)` loop that polls a volatile flag which signals which half of the DMA buffer is full, if any, and write that data to LFS.

The end of the loop will have an `osDelay()`, the duration of which needs to be no greater than `(time_to_fill_both_halves_of_the_DMA_buffer - time_to_write_a_half_buffer_to_LFS)`.

DeflateAwning:
Implemented in #384. Great!

PR #380, opened 2025-05-07
TCMD to setup camera, take picture, store in LFS

NuclearTea:
# Notes
- TCMD to setup camera to take a picture
- TCMD to capture camera
- Includes a DMA implementation for UART using 2 copy buffers (1 for halfCallback, 1 for CpltCallback)

## TODO
- [x] Remove Watchdog petting in `read_text_file` function in `lfs_telecommand`
- [ ] Remove Debug Statements
- [ ] Figure out what to do with `uart_telecommand_definitions`

PR #384, opened 2025-05-09
feat: Telecommand to capture image with camera

DeflateAwning:
Fixes Issue #8 

Replaces PR #380.

### DIFF
--- a/firmware/Core/Inc/camera/camera_capture.h
+++ b/firmware/Core/Inc/camera/camera_capture.h
@@ -1,0 +1,31 @@
+#ifndef INCLUDE_GUARD__CAMERA_CAPTURE_H
+#define INCLUDE_GUARD__CAMERA_CAPTURE_H
+
+#include <stdint.h>
+
+#define CAM_SENTENCE_LEN    67
+
+// Note: 23 sentences is too few. 100 sometimes is too few. 250 is a good number, I think.
+#define CAM_SENTENCES_PER_HALF_CALLBACK 250
+
+#define CAM_BYTES_TO_RECEIVE_PER_HALF_CALLBACK ((CAM_SENTENCE_LEN) * (CAM_SENTENCES_PER_HALF_CALLBACK))
+
+
+void CAM_repeated_error_log_message();
+
+
+typedef enum CAM_capture_status_enum {
+    CAM_CAPTURE_STATUS_TRANSMIT_SUCCESS,
+    CAM_CAPTURE_STATUS_WRONG_INPUT,
+    CAM_CAPTURE_STATUS_LFS_NOT_MOUNTED,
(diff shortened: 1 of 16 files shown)


MrBear101:
**Additional testing notes**

- 2 second timeout code that causes issues with capturing image has been commented out.
- Ali's code to deal with data remaining in buffer has been added to this code and images are now being written in their entirety.

DeflateAwning:
Problems with this (fixed some of them)

1. Attempts to write the remaining data twice. [fixed - switched from Ali's to my approach]
2. Logs a warning if deleting the file fails (absurd; no reason to expect the file should exist). Not sure why we're trying to delete it instead of just using the proper open flags to write overtop of it. [TODO]
3. Mostly-black extremely simple images which contain less than a full buffer of data result in an exit code 4 from the while loop. Should continue on to check if there's anything left in the buffers. [fixed]

DeflateAwning:
Can you please commit the decoder script to either `misc_tools/` or `hardware_emulators/` please? Would be good to track that in this repo. Ideally run the ruff autoformatter on it first.

MrBear101:
**Testing Notes May 12, 2025**

- Removed the LFS_delete_file() method from capture image so we don't get warnings when capturing an image with a new or existing file name. Below is a sample of the output when capturing an image
```
CTS1+camera_capture(imgs/d_mode1,d)!
0946689006000+0000002025_E [TELECOMMAND:DEBUG]: Parsed telecommand (len=36): 'CTS1+camera_capture(imgs/d_mode1,d)!'
0946689006000+0000002266_E [TELECOMMAND:DEBUG]: Executing telecommand from agenda slot 0, sent at tssent=0, scheduled for tsexec=0, logging to file: 'None'.
0946689006000+0000002281_E [TELECOMMAND:NORMAL]: 🚀 Executing telecommand 'camera_capture'.
0946689006000+0000002992_E [LFS:NORMAL]: Opened image file: imgs/d_mode1
From while loop, half 1 filled: @000002E1FF

Last sentence half 1: @00F902E1D1A10F000AB22AAC7C54EA78ACC0716C52A360E69869CB40CB68D95A


From while loop, half 2 filled: @00FA02E1D4

Last sentence half 2: @01F302E14614BA23F9888C551C0C6E14FF00249EC692A5A96EAE855FED4558C1
```
- Changed the write file flags from append to truncate so now when capturing an image with an existing name the file will be overwritten
- Tested capturing an image and resetting before capture is completed – after mounting file system everything is intact and if capture image was given an existing filename then the old file is untouched.
- Confirmed that directories do work with capture image but directories must be created beforehand using the `fs_make_directory` telecommand. If a directory exists then setting file name in camera_capture to `{dir_name}/{file_name}` will put the image in a file called {file_name} within the {dir_name} directory

MrBear101:
Also can confirm that nested directories work as well (again, provided they are created before hand using the appropriate telecommand)

PR #390, opened 2025-05-09
Send `LOG_message` calls over AX100 radio

### LINKED ISSUE #363: Logs: Sends logs over radio
See packet format in https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/blob/main/docs/Non-Critical_Notes/Radio_Packet_Format.md

Waiting for #360.

parker-research:
Fixes #363


firmware/Core/Inc/log/log.h
```diff
@@ -3,8 +3,8 @@
 
 #include <stdint.h>
 
-// Messages up to 256 characters
-#define LOG_FORMATTED_MESSAGE_MAX_LENGTH 256
+// Restricted by radio packet length
+#define LOG_FORMATTED_MESSAGE_MAX_LENGTH 185
```
JohnathanBurchill, line 7: Suggest adding an issue to ensure review of all LOG_message calls in the firmware to handle cases where messages are longer than LOG_FORMATTED_MESSAGE_MAX_LENGTH. LOG_message is used for the "available_telecommands" telecommand, for example, which already truncates the output. Not proposing a solution yet, but we should probably track this.

firmware/Core/Src/comms_drivers/comms_tx.c
```diff
@@ -27,23 +27,48 @@ uint8_t COMMS_downlink_tcmd_response(
     int32_t remaining_len = response_len;
     uint8_t response_seq_num = 1;
     while (remaining_len > 0) {
-        const uint16_t this_packet_len = (remaining_len > COMMS_TCMD_RESPONSE_PACKET_MAX_DATA_BYTES_PER_PACKET)
+        const uint16_t this_data_len = (remaining_len > COMMS_TCMD_RESPONSE_PACKET_MAX_DATA_BYTES_PER_PACKET)
             ? COMMS_TCMD_RESPONSE_PACKET_MAX_DATA_BYTES_PER_PACKET
             : remaining_len;
 
         packet.response_max_seq_num = max_seq_num;
```
JohnathanBurchill, line 34: max_seq_num is calculated as a difference of two #defs, but max_seq_num is an unsigned integral type. Suggest a check or #error directive if AX100_DOWNLINK_MAX_BYTES inadvertently gets set to a value smaller than COMMS_TCMD_RESPONSE_PACKET_MAX_DATA_BYTES_PER_PACKET. (See line 18 of this file)

firmware/Core/Src/log/log_sinks.c
```diff
@@ -76,10 +77,7 @@ void LOG_to_umbilical_uart(const char msg[])
 /// @return void
 void LOG_to_uhf_radio(const char msg[])
 {
-    // FIXME: replace with UHF radio transmission
-    // For debugging and prototyping, implement comms on a second UART
-    // Using the umbilical here messes up log messages on the umbilical
-    // LOG_to_umbilical_uart("TODO: replace this with a UHF RADIO transmission\n");
+    COMMS_downlink_log_message(msg);
```
JohnathanBurchill, line 84: Noting that the i2c bus transmit to the AX100 will timeout (currently after 100 ms). Is that the only bottleneck for the timing of this command?

firmware/Core/Src/log/log.c
```diff
@@ -68,7 +68,7 @@ static const uint8_t LOG_SEVERITY_MASK_DEFAULT = LOG_SEVERITY_MASK_ALL;
 // least one channel. It cannot be turned off. It is not included in the
 // array of sinks.
 static LOG_sink_t LOG_sinks[] = {
-    {LOG_SINK_UHF_RADIO, "UHF radio", LOG_SINK_OFF, LOG_SEVERITY_MASK_DEFAULT},
+    {LOG_SINK_UHF_RADIO, "UHF radio", LOG_SINK_ON, LOG_SEVERITY_MASK_DEFAULT},
```
JohnathanBurchill, line 71: Suggest leaving this sink OFF until the firmware is complete. Force people who want to test the radio to enable it by hand. This would be a failsafe for missing attenuators, antennas etc.

parker-research: Telecommand responses via AX100 are already enabled in main. Definitely a trade-off, but I'd prefer to start testing with as much of the firmware enabled as possible for majority of testing.

firmware/Core/Inc/log/log.h
parker-research: Do note that telecommand responses support multi-part outputs, but logs are currently restricted to this max length to fix in one packet (as long shall be short-ish, and we don't want to hog downlink with logs where possible).

Issue #397 opened.

firmware/Core/Src/comms_drivers/comms_tx.c
parker-research: Hadn't considered `#error` directives. Great point!

firmware/Core/Src/log/log_sinks.c
parker-research: Need to investigate queue behaviour of AX100 a bit more

parker-research:
Also worth mentioning - This piece of documentation may be helpful to review in parallel: https://github.com/CalgaryToSpace/CTS-SAT-1-OBC-Firmware/blob/main/docs/Non-Critical_Notes/Radio_Packet_Format.md

parker-research:
This is implemented to a satisfactory state imo. 

Biggest issue encountered was that the previous implementation of `DEBUG_uart_print_mixed_array()` did a massive stack-allocation of arrays, which caused stack overflows/hardfaults. That function is reimplemented with way smaller stack allocation now. This wasn't a big issue for the already-implemented telecommand response downlinks, as they had a massive stack allocated (8192 bytes); was an issue for smaller RTOS tasks which logged full-lenght log messages though.

Confirmed that downlinked log packets work. For example:

```
0000: eb a0 f0 c2 a2 8a 00 03 30 39 34 36 36 38 37 39
0010: 34 31 30 30 30 2b 30 30 30 30 30 30 32 31 38 34
0020: 5f 45 20 5b 45 50 53 3a 4e 4f 52 4d 41 4c 5d 3a
0030: 20 45 50 53 20 50 44 55 20 68 6f 75 73 65 6b 65
0040: 65 70 69 6e 67 20 
```

Downlinked telecommand responses also still work.

DeflateAwning (approved):

parker-research:
Going to merge this now so that it's in the main branch for testing with as they get working on it tomorrow. 

Has been tested a decent amount now, and even saw `fs_list_directory` command work (very cool)! 

Maybe a little more work to do in revising the log format; the timestamp is reasonably long atm

PR #396, opened 2025-05-10
ADCS: Create TCMD to erase ADCS SD file

### LINKED ISSUE #381: ADCS: Assorted issues from testing
Testing with the ADCS uncovered some bugs, all of which are now fixed:
- ADCS_pack_to_file_info_struct test failed.
  - This was broken by a later merge commit which modified this to remove an unnecessary conversion from MS-DOS time to Unix time. 
- adcs_reset() reset the OBC instead. Reset reason: INDEPENDENT_WATCHDOG.
  - We fixed this by removing the checkback since we don't expect to receive a response.
- The wrong wheel was enabled for the Y-wheel ramp test commissioning modes command.
  - This was "fixed", but we will need to get the actual mounting parameters from CubeSpace to verify it. 
- The newly-written (not merged to main) telecommand to disable SD logging on a specific log failed.
  - This was removed, because we can just use the command to set the SD log config to do this.
- Set Estimation Parameters differs in some parameters between CubeSupport and the Firmware Manual.
  - CubeSpace helped us with this. There was a new version of the Firmware Manual which we weren't told about, which fixes this issue and no others.
- When enabled using power control, CubeSense 1 and 2 keep turning on and off for some reason.
  - This is expected behaviour under cleanroom circumstances, according to CubeSpace.
- Saving an image to the SD card returns error 1, even when it was actually successful.
  - This issue was fixed. Note that even a successful send of the command may not save an image if it's already busy saving one.
- There's a missing commissioning mode for the X/Z Wheel Polarity Tests.
  - This issue was fixed.
- The commissioning telemetry logs for the two 3-Axis Control Commissioning modes were missing Commanded Wheel Speeds.
  - This issue was fixed.
- Downloading the index file resulted in a hard fault.
  - This function was completely reimplemented. It also fixed a known issue which meant that we couldn't know how long the index file was.
- Downloading an SD card file returned Error 12, because the file counter did not match its index in the file list.
  - We were expecting this, but couldn't prove that it would happen until testing with the real ADCS. Removing this error check, which had become unnecessary, fixed this.

There's also one thing which isn't yet fixed:
- There's no specific telecommand to delete files from the ADCS SD card. For performance reasons, this is a good idea.

DeemDeem52:
Closes #381

### DIFF
--- a/firmware/Core/Inc/adcs_drivers/adcs_commands.h
+++ b/firmware/Core/Inc/adcs_drivers/adcs_commands.h
@@ -149,5 +149,6 @@ uint8_t ADCS_disable_SD_logging();
 uint8_t ADCS_disable_peripherals_and_SD_logs_without_stabilisation();
 uint8_t ADCS_disable_peripherals_and_SD_logs_with_stabilisation();
 uint8_t ADCS_get_sd_card_file_list(uint16_t num_to_read, uint16_t index_offset);
+uint8_t ADCS_erase_sd_file_by_index(uint16_t file_index);
 
 #endif /* INC_ADCS_COMMANDS_H_ */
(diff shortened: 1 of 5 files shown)


DeflateAwning (approved):
Lgtm

Optional: explain what the "file index" means/where you get it from. If it's pretty obvious though, probably no need

DeemDeem52:
Tested and merged! Thanks Parker!

PR #412, opened 2025-05-13
Create telecommand: `fs_list_directory_json`

### LINKED ISSUE #405: Telecommand: List directory, as JSON, into response buffer
It'll be best to copy the `fs_list_directory` command to list the directory in a JSON format for display on the ground. Reading many log messages is a big tough, whereas the multi-part transfer format of the telecommand response system is ideal.

DeflateAwning:
Resolves #405


DeflateAwning:
Tested. Fixed a bug with concatenating the input folder path (which can come in many formats - with/without leading/trailing slashes) with the filenames. 

Could have refactored into a "make full path" function.

PR #416, opened 2025-05-15
Fix ADCS SD card function watchdog bug

### LINKED ISSUE #415: ADCS bug: Download SD file function triggers the watchdog when trying to download more than 75 files
Per the name. Since we have to repeatedly advance the file list read pointer and then wait some time, we are stuck there. 
Three notes:
- I don't seriously expect us to have more than 75 files on the SD card at a time (indeed, I don't really expect us to have more than 32 at a time), and we do plan to erase files after we download them. So this is, in principle, fine.
- There's no check that a file doesn't exist when we advance the pointer. I implemented this for the index file, so it's pretty easy to do so for the file download and erase functions, too (to prevent somebody accidentally typing index 77 instead of 7 and triggering the watchdog)
- The real fix, of course, is to pet the watchdog every 70 files or so. I'll implement that, too.

DeemDeem52:
Fixes #415

### DIFF
--- a/firmware/Core/Src/adcs_drivers/adcs_internal_drivers.c
+++ b/firmware/Core/Src/adcs_drivers/adcs_internal_drivers.c
@@ -34,7 +34,7 @@ uint8_t ADCS_i2c_send_command_and_check(uint8_t id, uint8_t* data, uint32_t data
             while (!ack.processed) {
                 ack_status = ADCS_cmd_ack(&ack); // confirm telecommand validity by checking the TC Error flag of the last read TC Acknowledge Telemetry Format.
                 num_ack_tries++;
-                if (num_ack_tries > ADCS_PROCESSED_TIMEOUT) {
+                if (num_ack_tries > ADCS_PROCESSED_TIMEOUT_MS) {
                     return 5; // command failed to process in time
                     // note: sending another telecommand when the first has not been processed
                     // will result in an error in the next telecommand sent
@@ -48,7 +48,7 @@ uint8_t ADCS_i2c_send_command_and_check(uint8_t id, uint8_t* data, uint32_t data
             return cmd_status; // if the HAL had an error, tell us what that is
         }
         num_checksum_tries++;
-    } while (ack.error_flag == ADCS_ERROR_FLAG_CRC && num_checksum_tries < ADCS_CHECKSUM_TIMEOUT);  // if the checksum doesn't check out, keep resending the request
+    } while (ack.error_flag == ADCS_ERROR_FLAG_CRC && num_checksum_tries < ADCS_CHECKSUM_TIMEOUT_MS);  // if the checksum doesn't check out, keep resending the request
 
     return ack.error_flag; // if the HAL was successful and the ADCS command had an error, tell us what it is
 }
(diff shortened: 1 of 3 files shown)


DeflateAwning (commented):
FYI - I have, for all other purposes, strongly opposed petting the watchdog inside telecommands. All other subsystems have avoided needing to.

I'll allow it here, as these are non-critical commands as far as I can tell, and it seems to be the least-friction path.

Can I suggest adding a global timeout inside any loop that pets the watchdog? For example, at the start of `ADCS_save_sd_file_to_lfs`, set `const uint32_t function_start_time = HAL_GetTick();`. Then, inside the loop, check `if (HAL_GetTick() - function_start_time > 60000)` (or whatever you deem a reasonable threshold), and return a timeout error code if that happens (plus clean-up like closing files). Just adds another layer of safety in case something logically goes run in these functions and they never exit. 

Do note that the EPS watchdog will power cycle EVERYTHING (OBC, ADCS, AX100, etc.) if its watchdog isn't pet for 5 minutes. I'd prefer not to rely on that generally though.

DeflateAwning:
> I estimate that it will take around 5-6 minutes to download

Can we test this? What durations have you observed?

It should be possible to transfer and save that in under 5 minutes for sure. `1 MiB / 100000 bits per second` = 83 seconds. Even at 2x slower than that, it should be possible. If not, can we request chunks of a file from the SD card? 

I'd love to go through it with you to see where we're bottlenecking. Are you keeping the LFS file open between chunks (hint: you should be)? I think you are right now

DeemDeem52:
Sorry, I miscalculated.

It takes about 4 ms to send the command and receive a given 20-byte packet. There are 1024 such packets in a block, so it takes 5 seconds to download 20 KiB. We have to wait 100 ms to get the download packet or else we get junk data, so it's really 5.1 seconds for 20 KiB. 

For 1 MiB, we need about 52 20-KiB blocks, so it works out to about 4.5 minutes. So this shouldn't be too bad.

DeemDeem52:
Update on this: we did some testing, and found that we could load a 1181962-byte file to the filesystem without any issues in 218496ms (so about 3.5 minutes). This covers our worst-case use case, so we should be good with the timeout I've added!

firmware/Core/Src/adcs_drivers/adcs_internal_drivers.c
```diff
@@ -48,7 +48,7 @@ uint8_t ADCS_i2c_send_command_and_check(uint8_t id, uint8_t* data, uint32_t data
             return cmd_status; // if the HAL had an error, tell us what that is
         }
         num_checksum_tries++;
-    } while (ack.error_flag == ADCS_ERROR_FLAG_CRC && num_checksum_tries < ADCS_CHECKSUM_TIMEOUT);  // if the checksum doesn't check out, keep resending the request
+    } while (ack.error_flag == ADCS_ERROR_FLAG_CRC && num_checksum_tries < ADCS_CHECKSUM_TIMEOUT_MS);  // if the checksum doesn't check out, keep resending the request
```
DeflateAwning, line 51: It seems weird to me that you're comparing a counter (`num_checksum_tries`) to a duration (`ADCS_CHECKSUM_TIMEOUT_MS`).

I suspect that this comparison went unnoticed before when that duration constant didn't have units in its name.

DeflateAwning (commented):
Just noticed one minor thing - not sure if it's an issue or not, but it seemed weird to me

firmware/Core/Src/adcs_drivers/adcs_internal_drivers.c
DeemDeem52: Oh, no -- that's my bad! That shouldn't have been changed to _MS in the first place!

DeflateAwning (approved):
Thanks for clarifying that! Units matter in engineering :)

Issue #420, opened 2025-05-16
LEOPS FSM: Start-of-mission/post-ejection/bootup sequence and operation state machine


DeflateAwning:
When the satellite boots up (e.g., on a workbench during integration into the launch vehicle, upon deployment from the launch vehicle, rebooted in space, after a crash/watchdog reset in space), it will enter the `CTS1_OPERATION_STATE_BOOTED_AND_WAITING` state. This issue describes a state machine of how the satellite deploys its antennas safely, with fallbacks if single-system failures occur.

## TL;DR

* When the satellite boots with the RBF tag installed, the antennas won't deploy. The radio will never transmit. All other systems are operational.
* When the satellite boots with the RBF tag removed (i.e., floating through space), the satellite will wait 30 minutes, then deploy its antennas, then move onto its nominal state with radio transmissions enabled. 
* When the satellite reboots in space, the RBF tag will be removed (obviously). The satellite has several bypass checks to avoid waiting 30 minutes 
    * These checks are: Are the antennas already deployed? Are uplinks happening? Has a "deployment successful" flag been written to the filesystem?
* After we've determined the satellite has deployed its antennas successfully, use telecommands to create a file containing "1" named `/bypass_deployment_and_enable_radio.txt` in the filesystem.

## Precautions
1. **Do not uplink to the satellite with its antennas stowed.** If the satellite receives a radio uplink in any state, it will assume its antennas are deployed and will enable radio operation.
2. Only change the RBF tag state when the satellite is powered off.
    * Reason 1: State transitions while powered on are untested/unvalidated.
    * Reason 2: Timed events are based on system uptime, not duration since RBF state change.
3. Obviously, when the RBF tag is removed, the satellite MUST NOT sit on a bench with its deployment switches deployed for more than 25 minutes! It will deploy its antennas. 
4. Do not create the `/bypass_deployment_and_enable_radio.txt` file until floating through space (antennas deployed, etc.). Even then, it's only helpful if ANTS fails (i.e., can't use it to check that antennas are deployed) and/or 

## States

* Global variable `enum CTS1_operation_state_enum_t CTS1_operation_state`:

```c
typedef enum {
    CTS1_OPERATION_STATE_BOOTED_AND_WAITING, // Entry point. Booted, and awaiting a trigger to deployment. Emulate a 30-minute countdown to antenna deployment.
[...]

PR #424, opened 2025-05-17
Implement bootup operation FSM (deploy antennas in space)

DeflateAwning:
See #420 for full description.


DeflateAwning:
Potential remaining TODO: If the EPS is in safety mode, I'm not certain it will get out of safety mode on its own. Need to confirm. May need to add an "if in safety mode and the battery is charged, exit safety mode" condition.

DeflateAwning:
Testing notes:

* If the LED blinks slowly, it may mean other systems (e.g., EPS) are failing to connect and timing out.
* Noticed bug: `AX100_uptime_at_last_received_kiss_tcmd_ms` assigned in wrong location. Fixing separately.

DeemDeem52 (approved):
I'm not 100% sure that I haven't missed anything, but I don't really see anything which would bear noting. There are a couple comments, though.

firmware/Core/Src/rtos_tasks/rtos_bootup_operation_fsm_task.c
```diff
@@ -0,0 +1,303 @@
...
+        LOG_SYSTEM_OBC,
+        LOG_SEVERITY_NORMAL,
+        LOG_SINK_ALL,
+        "CTS1 operation mode changed from %s to %s (%s).",
+        CTS1_operation_state_enum_TO_str(CTS1_operation_state),
+        CTS1_operation_state_enum_TO_str(new_mode),
+        reason_str
+    );
+    CTS1_operation_state = new_mode;
+}
```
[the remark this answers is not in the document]
DeflateAwning: > For this, mightn't it be better to maybe log something if CTS1_operation_state == new_mode so there isn't any confusion? (Perhaps it might also help us catch something if there's an odd edge case?)

The current implementation ends up calling `set_operation_mode_and_log_if_changed` to re-enter the current state on repeat in many cases (i.e., `new_state == current_state`). That's why it's like this. I think this is reasonably acceptable in state machines, as the "next-state" logic often has a loopback arrow to the current state, iirc.

> I also don't know if it makes sense to log what the CTS1 operation mode has been changed to after the operation state is set instead of before. I don't think this is something that can fail, though, so maybe it doesn't.

Great point! I'm going to update it to make it so `CTS1_operation_state = new_mode;` runs, then `LOG_message(...)` runs. That way, if changing into the `NOMINAL_WITH_RADIO_TX` state, it gets logged over the radio.

firmware/Core/Src/rtos_tasks/rtos_bootup_operation_fsm_task.c
```diff
@@ -0,0 +1,303 @@
...
+    if (file_size > 0) {
+        return 1; // File exists.
+    }
+    else {
+        return 0; // File does not exist or error.
+    }
+}
+
+/// @brief Power on EPS channel, check if all antennas have been deployed.
+/// @note This function returns the same result on failure as on "antennas not deployed". The distinction is not useful to the caller.
```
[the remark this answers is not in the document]
DeflateAwning: Thanks! Adding:

```
/// @return 1 if all antennas are deployed. 0 if not all antennas are deployed. 0 if an error occurs.
```

DeflateAwning:
Rebased just now.

firmware/Core/Src/rtos_tasks/rtos_bootup_operation_fsm_task.c
```diff
@@ -0,0 +1,306 @@
...
+        reason_str
+    );
+}
+
+/// @brief Check if the filesystem has the file "/bypass_deployment_and_enable_radio.txt".
+/// @return If the file exists, return 1. If it does not exist, or on failure, return 0.
+/// @note This function returns the same result on failure as on non-existence. The distinction is not useful to the caller.
+static uint8_t does_filesystem_have_bypass_deployment_and_enable_radio_file(void) {
+    // Check if the file exists in the filesystem.
+    const char* filename = "/bypass_deployment_and_enable_radio.txt";
```
NuclearTea, line 40: Shouldn't this be a global variable somewhere so that in the place where this file would be created, it would use this filename exactly?

firmware/Core/Src/rtos_tasks/rtos_bootup_operation_fsm_task.c
```diff
@@ -0,0 +1,306 @@
...
+        LFS_mount(); // Steamroll on error.
+    }
+
+    // Check if the file exists.
+    lfs_soff_t file_size = LFS_file_size(filename);
+    if (file_size > 0) {
+        return 1; // File exists.
+    }
+    else {
+        return 0; // File does not exist or error.
```
NuclearTea, line 52: Correct me if I'm wrong but in the rest of the codebase, we use a return of 0 to indicate success and a return of anything other than 0 as a failure. Unsure if we need to deviate from that general rule here

firmware/Core/Src/rtos_tasks/rtos_bootup_operation_fsm_task.c
```diff
@@ -0,0 +1,306 @@
...
+    if (failed > 0) {
+        // Log this as an error because failing communication with this thing is a big deal.
+        LOG_message(
+            LOG_SYSTEM_ANTENNA_DEPLOY,
+            LOG_SEVERITY_ERROR,
+            LOG_SINK_ALL,
+            "ANT_CMD_report_deployment_status() -> Error: %d",
+            failed
+        );
+        return 0; // Error.
```
NuclearTea, line 76: Note above comment regarding returning 0 as error

firmware/Core/Src/rtos_tasks/rtos_bootup_operation_fsm_task.c
DeflateAwning: While that convention is really effective for things like drivers, telecommand implementations, etc., it's not really easily transferable to these sorts of state machines and other autonomous high-level actions.

The reason is captured in the name of the function. The purpose of that function is to encapsulate the entirety of the logic of checking whether that file exists: `does_filesystem_have_<the bypass file>` and `have_all_antennas_deployed()`.

In the file case, if there's an error checking whether the filesystem has that file (e.g., a mounting error, a data corruption error, a file-exists-but-is-empty condition, etc.), it does not matter to the state machine what the cause is. The critical part is just that we indicate "the condition you're checking is not true". 

In the antenna deployment case, if there's an error checking, we're explicitly documenting that way assume NOT all antennas have been deployed. The logic of the state machine, which defines that errors are to be treated as "antenna not deployed", is implement in this single-purpose `static` function. Again, the state machine does not "care" about the error.

Issue #423, opened 2025-05-17
LittleFS Bug: Telecommand "fs_demo_write_then_read()" returns error code upon successful completion


aaryandeshpande:
fs_demo_write_then_read() telecommand successfully open/creates, writes to, reads from and closes file, but still returns an error code. The error code returned is typically some arbitrary high number. The current hypothesis is that the error code returned is just the size of the contents written into the file, which is stored in the variable "read_result" - HOWEVER this hypothesis is not validated.

PR #437, opened 2025-05-24
LFS: Fixed LFS return types to remove overflow situations

Saksham-P:
The changes are only made to functions that use internal LFS functions returning `lfs_ssize_t` type.

Internal LFS functions returning `int` type remain unchanged (they are `int8_t`)

### DIFF
--- a/firmware/Core/Src/littlefs/littlefs_helper.c
+++ b/firmware/Core/Src/littlefs/littlefs_helper.c
@@ -272,7 +272,7 @@ int8_t LFS_delete_file(const char file_name[])
 /// @param write_buffer - Pointer to buffer holding the data to write
 /// @param write_buffer_len - Size of the data to write
 /// @return 0 on success, 1 if LFS is unmounted, negative LFS error codes on failure
-int8_t LFS_write_file(const char file_name[], uint8_t *write_buffer, uint32_t write_buffer_len)
+lfs_ssize_t LFS_write_file(const char file_name[], uint8_t *write_buffer, uint32_t write_buffer_len)
 {
     if (!LFS_is_lfs_mounted)
     {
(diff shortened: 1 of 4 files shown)


DeflateAwning:
Is this related to an issue? Or a bug? As far as I can, some of these aren't bugs and don't really clarify details.

LittleFS error codes all fit in int8:
```
enum lfs_error {
    LFS_ERR_OK          = 0,    // No error
    LFS_ERR_IO          = -5,   // Error during device operation
    LFS_ERR_CORRUPT     = -84,  // Corrupted
    LFS_ERR_NOENT       = -2,   // No directory entry
    LFS_ERR_EXIST       = -17,  // Entry already exists
    LFS_ERR_NOTDIR      = -20,  // Entry is not a dir
    LFS_ERR_ISDIR       = -21,  // Entry is a dir
    LFS_ERR_NOTEMPTY    = -39,  // Dir is not empty
    LFS_ERR_BADF        = -9,   // Bad file number
    LFS_ERR_FBIG        = -27,  // File too large
    LFS_ERR_INVAL       = -22,  // Invalid parameter
    LFS_ERR_NOSPC       = -28,  // No space left on device
    LFS_ERR_NOMEM       = -12,  // No more memory available
    LFS_ERR_NOATTR      = -61,  // No data/attr available
    LFS_ERR_NAMETOOLONG = -36,  // File name too long
};
```

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -272,7 +272,7 @@ int8_t LFS_delete_file(const char file_name[])
 /// @param write_buffer - Pointer to buffer holding the data to write
 /// @param write_buffer_len - Size of the data to write
 /// @return 0 on success, 1 if LFS is unmounted, negative LFS error codes on failure
-int8_t LFS_write_file(const char file_name[], uint8_t *write_buffer, uint32_t write_buffer_len)
+lfs_ssize_t LFS_write_file(const char file_name[], uint8_t *write_buffer, uint32_t write_buffer_len)
```
DeflateAwning, line 275: On our functions, I prefer using absolute return sizes (e.g., `uint32_t`) instead of these aliases.

Saksham-P:
The error codes do fit in `int8_t`. But functions that return the number of bytes read/written, they need a return type of `int32_t`. I will make them absolute return size asap

Saksham-P:
Related Issue #423 

This fixes the fake error returns which are actually successful function calls

DeflateAwning:
This does not fix #423.

The `lfs_` functions return the positive number of bytes. The functions you're modifying generally don't. You need to look at the docstrings and ensure they're being implemented per the docstrings (or that the docstrings are being updated). The returns are guarded by `if (result < 0)`.

The bug is that the result in the read function is a positive number. Docstring says that's an error code, but it's actually success.

DeflateAwning (approved):
Thanks!

PR #438, opened 2025-05-24
Change all instances of GPS to GNSS in repo

NuclearTea:
## Note
Another PR will be made for IOC changes


DeflateAwning:
Depending on how hard it is, would be nice to make git realize that it's a file move/rename for those files that changed name. 

I think the nominal solution is `git mv <old> <new>` and then commit; not sure what the best way is after committing though. Maybe soft-reset or something. 

Let me know if that's tough/unreasonable though! :)

NuclearTea:
> Depending on how hard it is, would be nice to make git realize that it's a file move/rename for those files that changed name.
> 
> I think the nominal solution is `git mv <old> <new>` and then commit; not sure what the best way is after committing though. Maybe soft-reset or something.
> 
> Let me know if that's tough/unreasonable though! :)

Yep, was just looking into this. It's weird because some files were moved but some werent. Maybe changing the folder and the filename had something to do with it. I also had to change some of the formatting because for some reason, lots of the gps files were indented by 1 space so that could also be the issue

DeflateAwning (approved):
Goods to to me generally though, thanks!

NuclearTea:
This is closed and being replaced by: #440

PR #439, opened 2025-05-24
ADCS: Telecommand to exit bootloader

### LINKED ISSUE #436: ADCS: Telecommand expects an argument but there shouldn't be one in some cases
The ADCS command implementation has cases where we want to send no second argument bytes specifically in the generic telecommand for TC101. This input was not accepted.

DeemDeem52:
Fixes #436 

I had to create a telecommand to exit the bootloader because it's a bootloader command with zero parameters, and thus needs HAL_Master_Transmit instead of HAL_Mem_Write.

### DIFF
--- a/firmware/Core/Src/adcs_drivers/adcs_commands.c
+++ b/firmware/Core/Src/adcs_drivers/adcs_commands.c
@@ -1832,7 +1832,7 @@ uint8_t ADCS_erase_sd_file_by_index(uint16_t file_index) {
     return erase_status;
 }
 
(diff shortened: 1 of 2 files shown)


DeflateAwning:
Recommend pull request title: "ADCS Telecommand: Exit bootloader" (or similar) - it's not really clear what the bug is to me. Seems more like a missing feature.

DeflateAwning (approved):
Looks good, thanks!

Make sure to test it before merging, if not done already. Would be good to check that a few other commands work at least too!

PR #440, opened 2025-05-25
Change all instances of GPS to GNSS, using git mv

NuclearTea:
#435 

# Note
There will be an IOC file change that will be necessary. It will be in its own PR.

This will replace the current #438 PR as this has fewer line changed.


DeflateAwning (dismissed):
Ha GitHub is kinda stupid; each commit shows the moves/changes correctly, but the PR's entire diff doesn't seem to catch on. Onwards haha. Thanks for doing this!

DeflateAwning:
Just realized that docs were missed; should update in `docs/` (esp. the namespace one; maybe others too)

docs/C_Namespaces.md
```diff
@@ -28,7 +28,7 @@ following uppercase identifiers to indicate which subsystem/region of code it be
 * `ADCS_`: Attitude Determination and Control System
 * `EPS_`: related to the Electrical Power System
 * `EPS_CMD_`: functions which trigger a command to the EPS subsystem
-* `GPS_`: GPS/GNSS receiver
+* `GNSS_`: GNSS/GNSS receiver
```
DeflateAwning: Mid

DeflateAwning (approved):

docs/C_Namespaces.md
NuclearTea: .... That's annoying. MB

DeflateAwning: Ah oops sorry, I ended up deciding to just fix it because it was an easy mistake to make. Didn't mean to also comment on it :)

Issue #228, opened 2024-11-02
Add persistent logic to control dipole switch (coupled with ADCS and a backup/toggle system)


DeflateAwning:
* Mostly, pick the mode the ADCS claims to be set to.
* If it's been >20 minutes since a telecommand/communication, then start beaconing/flip-flopping between antenna states.

Probably put it in an RTOS thread (either new, or start grouping them).

Related to changes in #226.

PR #427, opened 2025-05-18
Create ADCS part and RTOS task part

DeflateAwning:
Resolves #228

PR #449, opened 2025-05-30
feat: RF Switch multi-mode control from ADCS, toggle-after-beacon, etc.

DeflateAwning:
Supercedes #427


DeflateAwning:
Fixes #228

Issue #344, opened 2025-04-23
Telecommand for GPS: Sync system time based on current GPS time


DeflateAwning:
Procedure for the telecommand:

1. Return error if power not enabled, probably.
1. GNSS send command get response (command: some command that gets the system time)
2. If invalid, return
3. Parse/extract out the time
4. Call the `TIM_set_current_unix_epoch_time_ms` function.

PR #451, opened 2025-05-30
Telecommand for GPS: Sync system time based on current GPS

meagan-davies:
For Issue #344


firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    // Extract UTC date/time components
+    // TODO: Might be better to use tokens[x] instead of tokens[count - y], Note: check if 8-13 works
+    int year       = atoi(tokens[14]);
+    int month      = atoi(tokens[15]);
+    int day        = atoi(tokens[16]);
+    int hour       = atoi(tokens[17]);
+    int minute     = atoi(tokens[18]);
+    int m_second   = atoi(tokens[19]); // TODO: Fix milliseconds.
```
DeflateAwning, line 71: Is todo still relevant?

Maybe a better variable name is `seconds_and_milliseconds_ms`

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    int count = 0;
+    char* token = strtok(copy_of_input_str, ",;*");
+    while (token != NULL && count < 25) {
+        tokens[count++] = token;
+        token = strtok(NULL, ",;*");
+    }
+
+    // Extract UTC date/time components
+    // TODO: Might be better to use tokens[x] instead of tokens[count - y], Note: check if 8-13 works
+    int year       = atoi(tokens[14]);
```
DeflateAwning, line 66: Prefer `int32_t` instead of `int` - be precise with types

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    if (!input_str) return 1;
+
+    // Create a safe, modifiable copy of the input
+    char copy_of_input_str[512];
+    strncpy(copy_of_input_str, input_str, sizeof(copy_of_input_str));
+    copy_of_input_str[sizeof(copy_of_input_str) - 1] = '\0';
+
+    // Tokenize input using both comma and semicolon as delimiters
+    char* tokens[25];  
+    int count = 0;
```
DeflateAwning, line 57: Prefer `uint8_t` instead of `int`

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * @brief Parses a GNSS receiver TIMEA message and converts it into Unix epoch time in milliseconds.
+ *
+ * The function expects a full TIMEA log string as input (e.g., starting with "#TIMEA,...").
+ * It tokenizes the string, extracts UTC date and time fields, validates their integrity,
+ * and converts them to a Unix timestamp in milliseconds.
+ *
+ * @param input_str The GNSS response string (e.g., from a TIMEA log).
+ * @return uint64_t Unix timestamp in milliseconds, or >0 on failure.
```
DeflateAwning, line 44: This return value doesn't make sense. The unix timestamp is >0 on success, and it "returns >0 on failure".

I'd recommend returning a success/failure code, and setting the value to a destination pointer passed in as an argument.

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+ */
+uint64_t GNSS_format_and_convert_to_unix_epoch(char* input_str) {
+    
+    if (!input_str) return 1;
```
DeflateAwning, line 48: `return 1;` means "the current time is 1970-01-01 T 00:00:01", I think. Need to check on the promises made here.

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    // Clean up any trailing newlines in the status field
+    char* newline = strchr(utc_status, '\n');
+    if (newline) *newline = '\0';
+
+    // Reject invalid UTC status
+    if (strcmp(utc_status, "VALID") != 0) {
+        return 1;  // UTC time is not valid
+    }
+
+    int second = m_second / 1000;
```
DeflateAwning, line 84: I'd personally just skip this. Set `t.tm_sec = 0`, then add on `m_second` at the end.

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    int day        = atoi(tokens[16]);
+    int hour       = atoi(tokens[17]);
+    int minute     = atoi(tokens[18]);
+    int m_second   = atoi(tokens[19]); // TODO: Fix milliseconds.
+    char* utc_status = tokens[20];
+    
+
+    // Clean up any trailing newlines in the status field
+    char* newline = strchr(utc_status, '\n');
+    if (newline) *newline = '\0';
```
DeflateAwning, line 77: Please use braces around bodies of if-statements

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    int month      = atoi(tokens[15]);
+    int day        = atoi(tokens[16]);
+    int hour       = atoi(tokens[17]);
+    int minute     = atoi(tokens[18]);
+    int m_second   = atoi(tokens[19]); // TODO: Fix milliseconds.
+    char* utc_status = tokens[20];
+    
+
+    // Clean up any trailing newlines in the status field
+    char* newline = strchr(utc_status, '\n');
```
DeflateAwning, line 76: These lines don't really make much sense tbh; can't tell if they're actively harmful though

firmware/Core/Src/unit_tests/test_gnss_time.c
```diff
@@ -0,0 +1,37 @@
...
+    struct tm t = {0};
+    t.tm_year = 2025 - 1900;
+    t.tm_mon  = 5 - 1;
+    t.tm_mday = 22;
+    t.tm_hour = 18;
+    t.tm_min  = 20;
+    t.tm_sec  = 28;
+
+    time_t expected_seconds = portable_timegm(&t);
+    uint64_t expected_ms = (uint64_t)expected_seconds * 1000 + 526;
```
DeflateAwning, line 30: Awesome to show this logic, but can you just use an online calculator to get a single hard-coded value instead of replicating this logic? Duplicating the logic from the function here means that, if the logic is wrong, it'll be wrong in both places.

I think do `const uint64_t expected_ms = 1747947628526;`

firmware/Core/Src/unit_tests/test_gnss_time.c
```diff
@@ -0,0 +1,37 @@
...
+
+#include <stdio.h>
+#include <time.h>
+#include <stdint.h>
+#include <string.h>
+
+/**
+ * @brief Unit test for GNSS_format_and_convert_to_unix_epoch
+ */
+uint8_t TEST_EXEC__GNSS_format_and_convert_to_unix_epoch() {
```
DeflateAwning, line 16: Please add at least 2 more test cases - one more in the year 2027, and two cases of invalid and partially-invalid strings.

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    copy_of_input_str[sizeof(copy_of_input_str) - 1] = '\0';
+
+    // Tokenize input using both comma and semicolon as delimiters
+    char* tokens[25];  
+    int count = 0;
+    char* token = strtok(copy_of_input_str, ",;*");
+    while (token != NULL && count < 25) {
+        tokens[count++] = token;
+        token = strtok(NULL, ",;*");
+    }
```
DeflateAwning, line 62: I think there's going to be an out-of-bounds access exception (very bad) if it reaches the end of execution and `count < 20` (e.g., if an invalid or partial string is passed here).

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+ * @brief Parses a GNSS receiver TIMEA message and converts it into Unix epoch time in milliseconds.
+ *
+ * The function expects a full TIMEA log string as input (e.g., starting with "#TIMEA,...").
+ * It tokenizes the string, extracts UTC date and time fields, validates their integrity,
+ * and converts them to a Unix timestamp in milliseconds.
+ *
+ * @param input_str The GNSS response string (e.g., from a TIMEA log).
+ * @return uint64_t Unix timestamp in milliseconds, or >0 on failure.
+ */
+uint64_t GNSS_format_and_convert_to_unix_epoch(char* input_str) {
```
DeflateAwning, line 46: Suggest rename: `GNSS_parse_timea_response_and_convert_to_unix_time_ms`

DeflateAwning (changes requested):
Thanks for working on this! Few pretty fundamental issues to resolve though - this has the potential to cause hard faults when run (e.g., if a partial string is passed to the parser function).

firmware/Core/Src/gnss_receiver/gnss_time.c
```diff
@@ -0,0 +1,165 @@
...
+    
+    if (!input_str) return 1;
+
+    // Create a safe, modifiable copy of the input
```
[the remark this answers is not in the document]
meagan-davies: The strtok to tokenize it modifies it as far as I understand

PR #461, opened 2025-05-31
LFS: Delete directory telecommand

Saksham-P:
The current method uses recursion, and isn't benchmarked for huge number of files and folders.


DeflateAwning (approved):
LGTM! Low blast radius, and nice to have.

PR #465, opened 2025-05-31
fix: Ensure camera EPS channel on test pass and/or failure

### LINKED ISSUE #364: Bug: Self-check leaves camera EPS channel enabled if camera test fails
Observed during MPI testing. Not reproduced, and involves a bit of speculation. Should check it though.

DeflateAwning:
Fixes #364


DeflateAwning:
Tested. All good. Checked with success and fail camera cases. Neat.

PR #467, opened 2025-06-01
MPI: Create `mpi_firmware_upgrade` telecommand

### LINKED ISSUE #18: MPI over-the-air firmware upgrade
### Telecommand: flash MPI with a specific MPI firmware file
* Arg 1: filename of new MPI firmware to flash

### Procedure

1. Power on the MPI
2. Send the bootloader command
3. Firehose dump the file over UART to the MPI (no checking for responses, currently)

<details>

### ~~Telecommand 1: uplink chunk of new MPI firmware~~ -> Already completed ( `CTS1+fs_write_file_hex` )
* Arg 1: start address into file of the MPI firmware
* Arg 2: up to 256 bytes of MPI firmware (in base64)
* Arg 3: filename of MPI firmware
* Arg 4: total filesize of MPI firmware

Function must write the incoming data to the "arg 2" byte address in a file in the filesystem.

### ~~Telecommand 2: check hash of new MPI firmware~~ -> Moved to Issue #341
* Arg 1: intended hash of new MPI firmware
* Arg 2: filename of new MPI firmware

</details>

DeflateAwning:
Resolves #18

Requires testing. Not certain it'll work perfectly.

### DIFF
--- a/firmware/Core/Src/telecommands/telecommand_definitions.c
+++ b/firmware/Core/Src/telecommands/telecommand_definitions.c
@@ -1092,6 +1092,7 @@ const TCMD_TelecommandDefinition_t TCMD_telecommand_definitions[] = {
         .number_of_args = 0,
         .readiness_level = TCMD_READINESS_LEVEL_FOR_OPERATION,
     },
+    {
         .tcmd_name = "mpi_send_firmware_upgrade",
         .tcmd_func = TCMDEXEC_mpi_send_firmware_upgrade,
         .number_of_args = 1,


JohnathanBurchill:
It is 11,11x bytes (I forget the 'x'). Around 10k bytes.

DeflateAwning:
Neat. Assuming 15,000 bytes, that's 30 chunks of 512 bytes. If we pause for 100ms between chunks, and each chunk transfers for about 20ms, then the upgrade should take about 3.6 seconds (easy).

DeflateAwning:
Discussion comments from today:
1. There's an "at the end of sending the file, you have to send the CRC and a command to exit". This should be covered by the generic send command get response deal.
2. There are two ways to upload firmware: the in-program way where you do the "enter bootloader" command (implemented here, but depends on non-bricked firmware), and also a way that sends a command in the first second upon bootup (more robust, but requires a bit of reverse engineering on the Atmel programmer mechanism).

PR #468, opened 2025-06-01
fix: Add chunking arguments/limits to `fs_read_` telecommands

### LINKED ISSUE #367: Bug: LFS read hex command triggers watchdog for large files
Documenting this so everyone is aware of it (though I vaguely recall that this one will be removed or modified before flight).

For sufficiently large files (I tested this with a 100 kB file from the ADCS CubeComputer SD card which was placed into the filesystem), calling the LittleFS reading command will cause the files to be continuously read, opened, and closed until for longer than the permitted 15 seconds from the watchdog.

To resolve this, either the function reading must be truncated or it must pet the watchdog periodically.

DeflateAwning:
Fixes #367


DeflateAwning:
Tested (mostly on small/short files). Seems good!

PR #477, opened 2025-06-01
Review: Log system telecommand response buffer output, mark `consts`, line lengths, return err codes

DeflateAwning:


DeflateAwning:
Also: Deleted `TCMDEXEC_log_report_latest_message_from_memory` because it always just reports "executing TCMDEXEC_log_report_latest_message_from_memory telecommand" (because that's always the log message immediately before it's executed)

Issue #464, opened 2025-05-31
MPI: Interleave timestamp data with MPI data collection


DeflateAwning:
### Background
In order to logically/analytically align the satellite's position with the MPI data, the GNSS data and the MPI data must both be timestamped.

GNSS data contains the timestamp data (in the format of "GPS reference week number, Seconds from the beginning of the GPS reference week accurate to the millisecond level").

### What to do
The MPI data must have the unix timestamp that the data was received interleaved periodically. It does not need to be written every frame, but a format must be defined to write the 64-bit unix timestamp (or perhaps the `sync_time+duration_since_sync` format) to the file, using the timestamp of the time that the buffer was marked as full (or the timestamp that the buffer started filling). 

### Timing Spec
To achieve 100ms accuracy, and assuming 5000 bytes per second nominal output, the timestamp must be placed +/- 500 bytes (or about 3-4 frames).

parker-research:
To quantify the "very very good" from previously, I think +/- 10ms would be a reasonable label. Perhaps +/- 2ms is fair. 

There are no sources of latency from when buffer fills to when the "filled timestamp" is set. Both happen in an ISR back-to-back. Thus, MPI timestamping inaccuracies are strictly dependent on general timestamping inaccuracies (e.g., system clock drift when not using the HSE). 

**Re: the original issue:** The solution is to interleave JSON messages containing timestamps with the core MPI data. The interleaved JSON may interject MPI data, but in such a way that perfectly removing the JSON allows full, perfect recovery of the MPI data.

PR #463, opened 2025-05-31
MPI Science Data handling - Version 2

Saksham-P:
- 160 byte DMA buffer
- 20480 byte Median buffer

PR #486, opened 2025-06-02
MPI science data collection via DMA

### LINKED ISSUE #14: Create "MPI active and recording data" mode
### Telecommands:
1. Turn MPI on and start recording
     * Arg: Run duration in seconds, from 1 sec to 20 minutes.
2. Turn off MPI early
     * Turns the MPI off immediately, if it's active currently

### Requirements, when active and recording
* Start a file with the "recording start time" timestamp.
* Collect the data and write it to the filesystem.
* ~~Monitor temperature (from EPS), and stop experiment early if too hot~~ (moved to #462)

### Thread
* Use a thread to do the actual UART-to-array-to-FS operation
* Use a shared configuration variable, where the telecommand can set the variable, and the thread operated if enabled

DeflateAwning:
Superseded #463. Full credit to @Saksham-P for the bulk of this implementation.

Resolves #14.

Related to #464, but might not be a complete solution for it.

### DIFF
--- a/firmware/Core/Inc/mpi/mpi_command_handling.h
+++ b/firmware/Core/Inc/mpi/mpi_command_handling.h
@@ -16,6 +16,7 @@ extern uint8_t MPI_science_file_can_close;
 extern uint8_t MPI_science_data_file_is_open;
 extern uint32_t MPI_science_data_bytes_lost;
 extern lfs_file_t MPI_science_data_file_pointer;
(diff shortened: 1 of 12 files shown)


DeflateAwning:
Tested pretty well. All seems good.

Sample data downlinked:

[mpi4.dat.zip](https://github.com/user-attachments/files/20559466/mpi4.dat.zip)

PR #495, opened 2025-06-03
ADCS: Show file_type enum values as strings in listing, convert to JSON

DeflateAwning:

PR #498, opened 2025-06-05
Fix SD download bug (add debug log message only though)

DeemDeem52:


DeemDeem52:
@DeflateAwning, could you test this with the real ADCS and share the log with me? If we can recreate the bug, that would be best.

DeemDeem52:
It's also possible that #495 might fix this anyways -- if it does, please let me know!

DeflateAwning (approved):
No problems with this. Not sure what the context or anything here is though. Can we close this? Should this log message call be merged?

DeemDeem52:
It shouldn't be merged. It was trying to fix a bug you were experiencing during the original testing with the satellite, but I don't think it's been encountered since I pushed #495.

PR #499, opened 2025-06-05
Create telecommand for SD download with CRC16 checksum

### LINKED ISSUE #492: ADCS: Add download_sd_file with CRC telecommand
May be a separate telecommand, or may be a modification of the original.

DeemDeem52:
Closes #492

### DIFF
--- a/firmware/Core/Src/adcs_drivers/adcs_commands.c
+++ b/firmware/Core/Src/adcs_drivers/adcs_commands.c
@@ -1570,9 +1570,11 @@ uint8_t ADCS_get_sd_card_file_list(uint16_t num_to_read, uint16_t index_offset) 
 /// @brief Save a specified file from the ADCS SD card to the ADCS subfolder in LittleFS.
 /// @param[in] index_file_bool Whether this is the index file or not
 /// @param[in] file_index Index of the file in the SD card; only used if index_file_bool is false
+/// @param[in] include_checksum_bool Whether to check the checksum or not
+/// @param[in] checksum CRC16 checksum of the file
 /// @return 0 if successful, non-zero if a HAL or ADCS error occurred in transmission, negative if an LFS or snprintf error code occurred. 
 /// Specifically, assuming no HAL or LFS error: bytes 0-2 are the ADCS error, bytes 3-10 are which command failed, bytes 11-16 are the index of the failure if applicable
-int16_t ADCS_save_sd_file_to_lfs(bool index_file_bool, uint16_t file_index) {
+int16_t ADCS_save_sd_file_to_lfs(bool index_file_bool, uint16_t file_index, bool include_checksum_bool, uint16_t checksum) {
 
     const uint32_t function_start_time = HAL_GetTick();
 
@@ -1637,6 +1639,12 @@ int16_t ADCS_save_sd_file_to_lfs(bool index_file_bool, uint16_t file_index) {
         const uint8_t file_info_status = ADCS_get_file_info_telemetry(&file_info);
         if (file_info_status != 0) {
             return file_info_status;
+        }
+
+        if (include_checksum_bool && (file_info.file_crc16 != checksum)) {
+            // check the CRC16 checksum
+            LOG_message(LOG_SYSTEM_ADCS, LOG_SEVERITY_ERROR, LOG_all_sinks_except(LOG_SINK_FILE), "CRC16 checksum incorrect at file index. (got %x)", file_info.file_crc16);
+            return 12;
         }
 
         // name file based on type and timestamp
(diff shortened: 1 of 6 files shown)


DeemDeem52:
@DeflateAwning It seems I've fallen pretty sick, so could this also be included for testing with the real ADCS?

firmware/Core/Src/adcs_drivers/adcs_commands.c
```diff
@@ -1585,9 +1585,11 @@ uint8_t ADCS_get_sd_card_file_list(uint16_t num_to_read, uint16_t index_offset)
 /// @brief Save a specified file from the ADCS SD card to the ADCS subfolder in LittleFS.
 /// @param[in] index_file_bool Whether this is the index file or not
 /// @param[in] file_index Index of the file in the SD card; only used if index_file_bool is false
+/// @param[in] include_checksum_bool Whether to check the checksum or not
```
DeflateAwning: Bad variable name imo. Is the meaning actually `enable_checksum_validation`?

DeflateAwning (changes requested):
At least the way I envisioned what was described in #492 (the issue related to this PR), the purpose wasn't an additional level of validation - it was that the checksum would be used to **identify** the file (as an alternative to using the index).

Iirc, the indexes change around a little bit if a photo is being taken in the background, if photos get deleted, etc. The idea behind using the checksum as the file identifier is that, even if the numerical IDs change, the sort order changes, etc., you can still request a specific photo. 

Consider the mission ops scenario of taking a photo on Jan. 1, but not deciding to downlink it until Mar. 1. On Jan. 2 (the day after taking the photo), you can note down the checksum, which will remain the same two months later - the IDs may change (I think). 

I'm not certain the IDs were changing, but it was a cause for confusion during testing in the cleanroom. We'll potentially have even more confusion during the mission.

DeemDeem52:
Everything is tested and works properly!

firmware/Core/Src/adcs_drivers/adcs_commands.c
```diff
@@ -1743,6 +1751,175 @@ int16_t ADCS_save_sd_file_to_lfs(bool index_file_bool, uint16_t file_index) {
...
+        switch(file_info.file_type) {
+
+            case ADCS_FILE_TYPE_TELEMETRY_LOG:
+                snprintf_ret = snprintf(filename_string, 17, "ADCS/log%d.TLM", file_index);
+                if (snprintf_ret < 0) {
+                    return snprintf_ret;
+                };
+                break;
+            case ADCS_FILE_TYPE_JPG_IMAGE:
+                snprintf_ret = snprintf(filename_string, 17, "ADCS/img%d.jpg", file_index);
```
DeflateAwning: Do these filenames actually make sense? Should they be named after the checksum instead?

firmware/Core/Src/telecommands/telecommand_adcs.c
```diff
@@ -1865,13 +1865,62 @@ uint8_t TCMDEXEC_adcs_download_sd_file(const char *args_str,
...
 
     return status;
 }
 
+/// @brief Telecommand: Download a specific file from the ADCS SD card by the index
+/// @param args_str 
+///     - Arg 0: The index of the file to download
+///     - Arg 1: CRC16 checksum of the file as two hex bytes in order (e.g. "07 ff" becomes 0x07ff)
+/// @return 0 on success, >0 on error
+uint8_t TCMDEXEC_adcs_download_sd_file_by_index_with_checksum(const char *args_str,
```
DeflateAwning: I'm not sure we need to add this new telecommand at all tbh. Simplicity in operation/explanation might be good here.

If we want to keep it, the docstring needs to explain [in some way] what happens in each of these cases:

1. Checksum argument doesn't match checksum of file with that index [I think it emits an error].
2. Checksum argument is incorrect and doesn't match the real checksum of the file [I think it fails]. 

The normal way checksums are used is by being sent alongside the data (e.g., by the index file), and then being compared/validated. It's strange it's an argument here imo - if we wanted to do data validation [which I don't think is that necessary], I'd sorta have expected the `ADCS_` functions to fetch it autonomously.

The other "normal" way checksums can be used is as a lookup key (as in "adcs_download_sd_file_by_checksum"), of course.

DeflateAwning (changes requested):
Really great! Just a few pedantic comments. Thank you! :)

DeflateAwning:
I think maybe some of the review comments got missed potentially (or feel free to respond if you think it's better another way). Looks pretty good now I think though. Thanks!

DeemDeem52:
I've fixed all but one of the requested changes (the last one is incoming in a bit) and rebased from main as well.

DeemDeem52:
Tested on the FlatSat, and worked as expected. Thanks Parker!

PR #501, opened 2025-06-11
Shortened log strings

celinahy:
Abbreviated the system and severity strings, implementing the suggestions and more. Not sure if this was what I needed to do, hence the draft PR.


DeflateAwning (approved):
Thanks

PR #505, opened 2025-06-14
Removed TCMD_TelecommandChannel_enum_t

celinahy:
Removed all instances of `TCMD_TelecommandChannel_enum_t`, and any references to it. Build passes.


celinahy:
> Assuming you did this with a regex, it's nice to include those in the commit messages so that it can be done again quickly if necessary.

I did not actually 😅, I felt like I had better control tracking down the uses from the .h files, which is why my commits are the way they are. I modified 3 ish .h files at a time and the corresponding .c file.

> I'm curious on a before/after removing this RAM and FLASH usage (as outputted by the linker at the end of the build).

Before:
Screenshot 2025-06-17 at 7 39 27 PM

After:
Screenshot 2025-06-17 at 7 38 27 PM

DeflateAwning (approved):
Approved on the basis of the last review being good, thanks!

DeflateAwning (changes requested):
Please rebase and ensure it builds before merging

DeflateAwning:
Rebased and fixed the new telecommand(s). Need this merged before it makes more work for other branches. Merging now.

PR #511, opened 2025-06-21
ADCS: Add adcs_bootloader_generic_command

DeemDeem52:
The ADCS bootloader doesn't support checksum. I've created a separate telecommand to send it commands, just in case we need any.

I don't expect that we will (the ones we might possibly need are already implemented), but without this telecommand it won't be possible to send other bootloader commands if we ever need them.


firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -734,6 +734,12 @@ const TCMD_TelecommandDefinition_t TCMD_telecommand_definitions[] = {
         .number_of_args = 2, 
         .readiness_level = TCMD_READINESS_LEVEL_FOR_OPERATION,
     },
+    {
+        .tcmd_name = "adcs_bootloader_generic_command",
```
DeflateAwning, line 738: Recommend rename to `adcs_generic_bootloader_command`

DeemDeem52:
Thanks Parker! 

As for testing, probably there should be at least two tests -- one with arguments and one without. Performing an ADCS reset (ID 1) should take care of the former -- and as you suggested, the exit bootloader command (ID 101) should take care of the latter.

DeemDeem52:
(Had to rebase to take care of a conflict)

Issue #507, opened 2025-06-19
Telecommands: Bulk uplink a file while keeping it open between telecommands


DeflateAwning:
Create the following telecommands:
* `bulk_uplink_open_file(filename, truncate or append mode)`
* `bulk_uplink_close_file()` [no arguments]
* `bulk_uplink_write_bytes_hex()` [one argument - the bytes as hex]

Similar in concept to the bulk file downlink, but instead for uplinking a file (e.g., MPI firmware, new version of the OBC firmware, etc.).

DeflateAwning:
Critical sub-feature of #25

PR #514, opened 2025-06-29
TCMD to write the contents of a given file to the internal flash

NuclearTea:


DeflateAwning:
This looks great! Maybe we can write a sample sequence of how to test this in the `docs/` folder, and can do a bit more testing.

I think ultimately to be useful, it's dependent on #507 (bulk file uplink). I'm going to get working on that one right away too.

firmware/Core/Inc/stm32/stm32_internal_flash_drivers.h
```diff
@@ -25,11 +31,24 @@ typedef struct
...
+typedef enum
+{
+    STM32_INTERNAL_FLASH_WRITE_SUCCESS,
+    STM32_INTERNAL_FLASH_WRITE_ADDRESS_TOO_LOW,
+    STM32_INTERNAL_FLASH_WRITE_ADDRESS_TOO_HIGH,
+    STM32_INTERNAL_FLASH_WRITE_ADDRESS_OVERLAPS_BOTH_FLASH_BANKS,
+    STM32_INTERNAL_FLASH_WRITE_UNLOCK_FAILED,
+    STM32_INTERNAL_FLASH_WRITE_LOCK_FAILED,
+    STM32_INTERNAL_FLASH_WRITE_OPERATION_FAILED,
+} STM32_Internal_Flash_Write_Return_t;
```
DeflateAwning: Just in case you miss my pedantic reviews at all, I would have previously required this be renamed to `STM32_internal_flash_write_return_t` ;)

firmware/Core/Src/telecommands/telecommand_definitions.c
```diff
@@ -365,6 +365,12 @@ const TCMD_TelecommandDefinition_t TCMD_telecommand_definitions[] = {
         .number_of_args = 2,
         .readiness_level = TCMD_READINESS_LEVEL_FOR_OPERATION,
     },
+    {
+        .tcmd_name = "fs_write_file_to_internal_flash",
```
DeflateAwning: Maybe "stm32_internal_flash_write_from_file" - my fear is that all the other "fs_" commands are safe to use, but this one requires a giant warning label.

DeflateAwning (changes requested):
Just a few change requests about namespacing. 

Looks very awesome, thanks! 

I'd be happy/excited to include this with just a few more changes. Would be great to test first, but honestly, not much harm can be done by including it in the firmware even without a ton of testing.

firmware/Core/Inc/stm32/stm32_internal_flash_drivers.h
NuclearTea: This is probably a vestige of the coding format we use at work (Pascal case lol).

DeflateAwning:
Moved this into PR #611 to merge it as a contingency-type feature. Thanks again for working on it!

PR #515, opened 2025-07-21
Rewrite flash memory driver

Makiv1:
## Key Changes

* Fixed bug where CS line wasn't being pulled inactive (high) between operations, and thus writes weren't being committed/finished. Symptom: stall.
* Increased write speed by using SPI DMA access.

## Motivation

The old flash drivers had some bugs which were causing the file system to brick the obc when I tried to implement logging to the flash module.

### Some of the issues I observed:
1. There exists code paths where the chip select line is not pulled high before the driver function returns. This was the main culprit causing my issues I think as it was causing the chip select to remain low (the active state) which could cause issues in subsequent driver calls. 
2. The flash write enabled bit was not zeroed before driver functions returned.
3.  `while (1)` loops were used which Is just an anti pattern. If something weird happens in the loop you brick the program.

### To remedy this I have rewritten the flash drivers with the following goals:
1. move handling of the chip select line into concise functions which handle all spi transactions
2. move common operations into functions so one can better reason about the workings of the code
3. ensure clean up is properly done for operations which require it

### The testing I've done so far:
- [x] writing log messages to a single file ( ~1.25 megabytes) with the occasional power cycle. Syncs are initially quick (<100ms) but slow down as the file grows (~5 sec)
image

- [x] Test writing to multiple files 

Notably I have not observed any crashes/bricking/filesystem corruption/watchdog timeouts with the new drivers, something that was quiet easy to replicate with the old ones (just `git checkout 93dae66b3be8ad80aed95dfe9a96c1881c2b66eb` and leave the code logging for a while) 

### Notes For The Reviewer:
1. Start with the `spi_internal` files
2. Then review the `flash_driver` files
3. Then the `lfs_driver files`
4. Finally everything else

### DIFF
--- a/firmware/Core/Src/littlefs/flash_driver.c
+++ b/firmware/Core/Src/littlefs/flash_driver.c
@@ -16,7 +16,7 @@ FLASH_error_enum_t FLASH_erase_block(uint8_t chip_number, FLASH_Physical_Address
     FLASH_write_enable(chip_number);
 
     // Send erase command along with the address of the block.
-    uint32_t block_address = address.row_address; // Address to the start of the block.
+    const uint32_t block_address = address.row_address; // Address to the start of the block.
     uint8_t cmd_buff[] = {FLASH_CMD_BLOCK_ERASE, ((block_address >> 16) & 0xFF), ((block_address >> 8) & 0xFF), (block_address & 0xFF)};
     FLASH_SPI_Data_t cmd = {.data = cmd_buff, .len = sizeof(cmd_buff)};
 


Makiv1:
Took a further look at the flash functions and I found the bottle neck. It's the `HAL_spi_transmit` and `HAL_spi_receive` calls.
The SPI1 bus has a baud rate of 2Mbits/sec. Transfering a page (2048 bytes or 16.4 kilobits) at this baud rate would take 
$$\frac{16.4 kilobits * 1 sec}{2000 kilobits}=0.0082 seconds = 8.2milliseconds$$ which explains the benchmark above. 

Setting the clock pre-scaler from 8 to 2 gives an effective baud rate of 8Mbits/sec which is  2.05 milliseconds per page. This gives the following benchmarks:
image

Reads are still far to long. I suspect this is an issue with `HAL_Receive` adding per byte overhead. Will try  using DMA mode.

In general erasing, programming and reading should all be sub 1 millisecond. But if we want those speeds we will need to increase the clock speed of the SPI1 bus.

Makiv1:
Using DMA when receiving brings down the cost of page reads to ~3ms. At this point flash operations are only bottle necked by by the SPI clock speed.
image

Makiv1:
#### On the topic of file closes and syncs
closes and sync are relatively cheap (~30ms), until they are not... (~8sec)

the massive spike it time happens when metadata compaction is triggered by lfs. see https://github.com/littlefs-project/littlefs/issues/1079#issuecomment-2720048008 for an explanation.

In order to bound the time taken by metadata compaction we must set `.metadata_max` to a low enough value. Otherwise when the filesystem is large enough, the watchdog will trigger every time lfs tries to perform metadata compaction.

firmware/Core/Src/littlefs/littlefs_helper.c
```diff
@@ -37,27 +34,44 @@ struct lfs_config LFS_cfg = {
     .prog_size = FLASH_CHIP_PAGE_SIZE_BYTES,
     .block_size = FLASH_CHIP_BLOCK_SIZE_BYTES,
     .block_count = (FLASH_CHIP_SIZE_BYTES / FLASH_CHIP_BLOCK_SIZE_BYTES),
-    .block_cycles = 100, // TODO: ASK ABOUT THIS (HOW FREQUENT ARE WE USING THE MODULE),
+    .block_cycles = 500, // TODO: ASK ABOUT THIS (HOW FREQUENT ARE WE USING THE MODULE),
```
DeflateAwning: Should resolve TODO, or leave note about how to resolve it/what this field means.

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -1,871 +1,226 @@
...
-    FLASH_activate_chip_select(chip_number);
-    const HAL_StatusTypeDef tx_result_1 = HAL_SPI_Transmit(hspi, (uint8_t *)&FLASH_CMD_GET_FEATURES, 1, FLASH_HAL_TIMEOUT_MS);
-
-    // If couldn't send the command, check why
-    if (tx_result_1 != HAL_OK) {
-        FLASH_deactivate_chip_select();
+FLASH_error_enum_t FLASH_program_page(
+    uint8_t chip_number, FLASH_Physical_Address_t address, uint8_t *data, uint32_t data_len
+) {
+    // FLASH_disable_block_lock(chip_number); // TODO: why do we need to disable block lock? it's never turned back on?
```
DeflateAwning: Can probably just drop this comment. It's unlocked on reboot, so if there's ever issues with it getting locked, a reboot should fix it?

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -1,871 +1,226 @@
...
-    const FLASH_error_enum_t wren_result = FLASH_write_enable(hspi, chip_number);
-    if (wren_result != FLASH_ERR_OK) {
-        FLASH_deactivate_chip_select();
-        return wren_result;
-    }
-
-    // Send PROGAM LOAD Command
-    FLASH_activate_chip_select(chip_number);
-    const uint8_t tx_result_1 = HAL_SPI_Transmit(hspi, (uint8_t *)&FLASH_CMD_PROGRAM_LOAD, 1, FLASH_HAL_TIMEOUT_MS);
+    // TODO: something like this is probably worth having.
```
DeflateAwning: Should we add it back then?

firmware/Core/Src/littlefs/littlefs_helper.c
DeflateAwning: Related to how active your wear-leveling is. 500 seems good. Can remove the TODO.

DeflateAwning:
Seems like a solid performance improvement! Did some testing with the MPI emulator script.

```
# With current `main` branch:
MPI Task: Successfully wrote 20480 bytes to file in 355ms

# With this branch:
MPI Task: Successfully wrote 20480 bytes to file in 133ms
```

Another interesting observation, which likely makes sense, is that a filesystem formatted with the old firmware cannot be mounted with this version (likely due to parameter reconfiguration). The following error code is raised on attempting to mount:

```
    LFS_ERR_INVAL       = -22,  // Invalid parameter
```

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -1,871 +1,226 @@
...
-    // If couldn't send the command, check why
-    if (tx_result_1 != HAL_OK) {
-        FLASH_deactivate_chip_select();
+FLASH_error_enum_t FLASH_program_page(
+    uint8_t chip_number, FLASH_Physical_Address_t address, uint8_t *data, uint32_t data_len
+) {
+    // FLASH_disable_block_lock(chip_number); // TODO: why do we need to disable block lock? it's never turned back on?
+    FLASH_write_enable(chip_number);
+    
+    // TODO: not sure if I'm doing the memory mapping correctly. maybe should be offset as the address?
```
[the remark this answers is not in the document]
Makiv1: resolved :)

Makiv1:
I think this is good to merge at this point!

firmware/Core/Src/littlefs/flash_driver.c
```diff
@@ -17,9 +16,9 @@ FLASH_error_enum_t FLASH_erase_block(uint8_t chip_number, FLASH_Physical_Address
     FLASH_write_enable(chip_number);
 
     // Send erase command along with the address of the block.
-    const uint32_t block_address = address.block_address;
```
DeflateAwning: Why did you remove `const`? Non-mutated variables should be marked as const.

DeflateAwning (approved):
Thanks

PR #517, opened 2025-07-27
Implement logging to file system

### LINKED ISSUE #290: LFS: Write logs to log file in LFS
Could setup a periodic job that copies from the in-memory circular buffer to LFS if LFS is too slow.

Otherwise, just need to set up the write on every `LOG_message` call to write to LFS.

Makiv1:
fixes #290 

Uses the rewritten flash drivers.

Bedtime now.


DeflateAwning:
Thanks for the work on this - it's all been moved to #533. Closing this PR because #533 covers everything here.

Issue #331, opened 2025-04-12
LFS Telecommand: Get byte offset of the nth string occurrence in a file


DeflateAwning:
Useful for locating and downlinking a specific warning/error message from log files.

Likely also useful for other things.

Ideally create an "occurrence of string" and "occurrence of hex pattern" version

PR #523, opened 2026-01-11
LFS Telecommands: Search LFS files for str/hex patterns, count occurrences, get byte offsets

### LINKED ISSUE #330: LFS Telecommand: Count occurrences of string in file
Useful for counting the number of warnings/errors in LFS log files, for example.

Ideally create an "occurrence of string" and "occurrence of hex pattern" version

DeflateAwning:
Resolves #330 and #331

### DIFF
--- a/firmware/Core/Src/littlefs/littlefs_searching.c
+++ b/firmware/Core/Src/littlefs/littlefs_searching.c
@@ -6,7 +6,7 @@
 #include "littlefs/littlefs_helper.h"
 
 /// @brief Count the number of occurrences of a string/byte pattern in an LFS file.
-/// @param filename File name/path to open.
+/// @param filename File name/path of the "haystack" file to open and search within.
 /// @param needle Pattern to search for.
 /// @param needle_len Length of pattern to search for.
 /// @return The number of matches. 0 if no matches. Negative LFS error code on error.
@@ -66,7 +66,7 @@ int32_t LFS_search_count_occurrences(
 
 /// @brief Find the byte offset of the nth occurrence of a string/byte pattern in an LFS file.
 /// @param lfs Filesystem.
-/// @param filename File name/path to open.
+/// @param filename File name/path of the "haystack" file to open and search within.
 /// @param needle Pattern to search for.
 /// @param needle_len Length of pattern to search for.
 /// @param n Nth occurrence to find.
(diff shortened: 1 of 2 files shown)


firmware/Core/Inc/telecommands/lfs_search_telecommand_defs.h
```diff
@@ -0,0 +1,32 @@
+#ifndef INCLUDE_GUARD__LFS_SEARCH_TELECOMMAND_DEFS
+#define INCLUDE_GUARD__LFS_SEARCH_TELECOMMAND_DEFS
```
DeemDeem52: Is there any particular reason that this is a separate file from the other LittleFS telecommands?

firmware/Core/Src/telecommands/lfs_search_telecommand_defs.c
```diff
@@ -0,0 +1,222 @@
...
+        strlen(arg_needle)
+    );
+
+    if (count < 0) {
+        snprintf(response_output_buf, response_output_buf_len,
+                 "Search error: %ld", count);
+        return 3;
+    }
+
+    snprintf(response_output_buf, response_output_buf_len, "%ld", count);
```
DeemDeem52: Maybe it's better to echo back (part of) the string search in some way like "Count of string %s in file %s: %ld" (obviously not using %s), to have some context? That might be a little too long, though.

DeemDeem52: (Just thinking of the case where this gets logged with a delay for some reason and then it's not clear what is being searched for)

DeemDeem52 (commented):
I'm not really noticing anything squiffy codewise... just a couple comments.

firmware/Core/Src/littlefs/littlefs_searching.c
```diff
@@ -0,0 +1,136 @@
...
+#include <stdlib.h>
+#include "littlefs/lfs.h"
+
+#include "littlefs/littlefs_searching.h"
+#include "littlefs/littlefs_helper.h"
+
+/// @brief Count the number of occurrences of a string/byte pattern in an LFS file.
+/// @param filename File name/path to open.
+/// @param needle Pattern to search for.
+/// @param needle_len Length of pattern to search for.
```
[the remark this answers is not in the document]
DeflateAwning: Yeah, the analogy is that you're searching a "needle in a haystack". I suppose that's lost when the other argument is "filename" instead of "haystack_filename" though.

Updated the comments to reflect this a bit more clearly, good catch, thanks!

firmware/Core/Src/telecommands/lfs_search_telecommand_defs.c
DeflateAwning: Great catch, thanks! That was sloppy on my part

PR #565, opened 2026-03-14
EPS: Add a `eps_get_pdu_housekeeping_data_eng_short_json` tcmd, use in logs

### LINKED ISSUE #557: `EPS_log_pdu_json` logs are truncated
Not very essential, and thus not critical to fix, but the logs in `EPS_log_pdu_json()` are truncated. Should make a shorter version of that PDU log message.

parker-research:
Resolves #557 

Note that this PR removes EPS PDU periodic logging (because it overran log buffer). The applicable PDU commands can still be scheduled in order to fetch that data though.

### Test Plan

1. Run the `eps_get_pdu_housekeeping_data_eng_short_json` telecommand
2. Observe the periodic logs
3. Try enabling a few channels and make sure the periodic logs don't get too long


parker-research:
All tests successful. Confirmed that nominal "low power" mode is around 77 cW input to the PDU.

PR #593, opened 2026-03-18
ADCS: Auto disable ADCS power when EPS enters safety mode

### LINKED ISSUE #584: Automatically disable ADCS power controls when EPS enters safety mode
The ADCS power channels from the EPS are always on (they come through the main Stack power). 

We must write a background task/subtask that checks the EPS state, and if it ever goes off-nominal, we must disable the ADCS power via its internal power controls.

Configuration variable for how often to check (or 0 to disable this feature).

Set all ADCS power controls to disabled if the EPS ever enters safety mode (controlled by its own internals).

Testing: Enable some ADCS power channels. Force the EPS into Safety mode, then watch as the subtask sets the ADCS to disabled.

Supercedes #37.

parker-research:
Resolves #584

### DIFF
--- a/firmware/Core/Inc/eps_drivers/eps_types.h
+++ b/firmware/Core/Inc/eps_drivers/eps_types.h
@@ -75,9 +75,16 @@ typedef struct {
     int16_t curr_ou_mppt_mA;
 } EPS_conditioning_channel_short_datatype_eng_t;
 
+typedef enum {
+    EPS_MODE_STARTUP = 0,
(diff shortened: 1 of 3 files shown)


parker-research:
Testing successful.

Rough procedure:

1. Set run mode.
2. Enable some channel powers.
3. Set EPS to safety.
4. See that OBC turns ADCS power controls off.
5. Switch EPS to nominal mode.
6. Enable ADCS power controls again.
7. Confirm they stay on for several moments

PR #605, opened 2026-03-19
GNSS: Skip null bytes in input stream in ASCII mode

parker-research:
Related to #596. Original implementation has the potential for a slight regression, which is fixed here.

### Summary of Changes

* GNSS: Skip null bytes in input stream in ASCII mode

* Fix: Fill missing beacon field (gnss_rx_mode_enum)

* Move the remove_null_bytes_in_middle step, add hex version of command


DeflateAwning (approved):

parker-research:
Testing is good. Tested:

* Time sync procedure (with the real satellite though - so it fails as expected, but still exits quickly)
* Receiving hex and ascii via the commands

Will test firehose storage mode in integration.

PR #614, opened 2026-03-20
Telecommand: obc_set_stm32_sysclk_to_hse (switch to HSE external clock)

DeflateAwning:
### Changes

* IOC file change: Peripherals (e.g., I2C, UART) now use `HSI` directly instead of `PCLK1` (which depends on SYSCLK speed). Benefit: We don't have to re-init all the peripherals.
* Created telecommand `obc_set_stm32_sysclk_to_hse`, which re-inits the clocks to use the HSE clock.
* Created telecommand `demo_os_delay`, for testing the osDelay time.

### Basic Test

1. Bootup.
2. Sync time.
3. Run `obc_set_stm32_sysclk_to_hse`
4. Run `demo_blocking_delay(10 seconds)` and confirm exact runtime with computer-side timestamped log messages.
4. Run `demo_os_delay(10 seconds)` and confirm exact runtime with computer-side timestamped log messages.
5. Do time sync to computer.
6. Wait minutes.
7. Do time sync to computer again to confirm minimal drift.
8. Run `demo_blocking_delay(100 seconds)` to ensure watchdog still triggers.

### Risks

* The "IOC Prework" commit is the highest risk with this change, as we can't explicity test 100% of the systems affected by it. That is, I2C is used to transmit on the radio (the same bus that goes through the stack and to the ADCS), and we can't test the radio with the antenna stowed. However, we can confirm that the stack I2C still works by using the ADCS on that channel (which I've done, for confirmation).
* Otherwise, this is a low-risk change, as the bootup is the same as the existing config, and we don't have to use the command until later on in the mission anyway. Watchdogs and other safety mechanisms (the "reset if no uplink recently" feature) also help mitigate any risks.

### DIFF
--- a/firmware/Core/Src/telecommands/obc_systems_telecommand_defs.c
+++ b/firmware/Core/Src/telecommands/obc_systems_telecommand_defs.c
@@ -132,7 +132,7 @@ static uint8_t SystemClock_Config_with_hse(void)
   }
 
   DEBUG_uart_print_str("HAL_PWREx_ControlVoltageScaling() successful.\n");
-  HAL_Delay(1500);
+  HAL_Delay(500);
   DEBUG_uart_print_str("Waited.\n");
 
   /** Initializes the RCC Oscillators according to the specified parameters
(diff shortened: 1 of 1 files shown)


DeflateAwning:
Discussed with @JohnathanBurchill briefly online. AI-assisted code review suggests this change is a reasonable approach. Will plan to forward with it, knowing a reversion is reasonably easy.

PR #663, opened 2026-07-28
Blob: Extended Beacon v2

parker-research:
This blob defines a new extended beacon format, with many more ADCS, EPS, etc. data fields.

When this blob is run, it downlinks a packet of this new beacon format, and then optionally reschedules itself for a few seconds in the future, allowing it to beacon repeatedly forever (until a reboot).


DeflateAwning (approved):
Great!

PR #665, opened 2026-08-09
Blob: Extended Beacon Blob v3

parker-research:
Changes from v2:
* disable log message at start of execution
* fix uplink counter to avoid incrementing on every beacon
* bump version


parker-research:
Deployed and working well!

DeflateAwning (approved):

